PR #110, opened 2014-10-04
Lily

### LINKED ISSUE #32: Interface to a general LTL-game solver.
(Copied from https://sourceforge.net/p/tulip-control/feature-requests/22/ with creation date 2012-05-31, by @necozay and assigned to Nick Spooner)

This will enable to do synthesis with specs beyond GR(1).

johnyf:
Adds interface to the full LTL synthesis tool [`Lily`](http://www.iaik.tugraz.at/content/research/design_verification/lily/) and thus close #32. The available choices of LTL solvers are listed [here](https://github.com/johnyf/tool_lists/blob/master/verification_synthesis.md#full-ltl-games) (so excluding parity game solvers for this).

From these:

- `GAVS+` is `Java`, GUI-oriented, and a bit too feature-full
- `GASt` is not really open-source
- `Unbeast` has obfuscated source code
- `Acacia` and `Alaska` have been superseded by `Acacia+`

so these are not candidates.

The remaining contenders are `Acacia+` and `Lily`. `Acacia+` has the advantages that:

- it is `python` on the surface
- currently maintained.

But it:

- includes `C`, so it needs compilation
- requires (a patched) `Lily`, so it is at least as difficult to install as `Lily`.

For this reason, as a first choice `Lily` was selected, even though it is the oldest (apparently the first LTL synthesizer) and recently updated. It's installation is easy and cross-platform, because it is written in `Perl`.


slivingston:
@johnyf , given your reasoning about tool selection in the opening post, I would expect that Lily is open source. However, I have found no indication of this in the current Lily release (the tarball name is [lily-2008-06-19.tar.gz](http://www.iaik.tugraz.at/content/research/design_verification/lily/lily-2008-06-19.tar.gz)). Indeed, I did not find any declaration of license, copyright, or dedication to the public domain. I also searched for license information in [Wring 1.1.0](ftp://vlsi.colorado.edu/pub/Wring-1.1.0.tar.gz) but found none. Do you know whether these tools are free as in freedom?

johnyf:
No, I am not aware of a license declaration for Lily.

slivingston:
It would be useful if someone who has Lily can try running the tests and example for the current interface with Lily. Do this by installing from the tip of `lily` branch (8aee6b8546ba5356b841aaa534266b94428853c7 at time of writing) and then
```shell
./run_tests.py lily
python examples/developer/interfaces/lily-example.py
```
Once someone confirms this, I will merge and close this pull request.

The interface has the following known fragilities:
- a syntax error during the invocation of `lily.pl` will result in an empty DOT file, which is not currently explicitly handled and thus leads to a cryptic error message from `networkx` when it tries to parse the file;
- The boolean constants `True` and `False` are not supported during the translation to Wring syntax, and I am not aware that expressing such constants directly is possible in the Wring syntax; cf. the LTL syntax definition of `LTL.pm` in Lily or Wring.

I recommend these be addressed after merging because good, robust solutions may not be arrived at until more experience has been gained with Lily and other LTL solvers in the hybrid systems contexts of TuLiP.

johnyf:
As of bab4d84336533e689886c7218bcfbdc9bc519e04 the `lily` tests pass and the example returns a Mealy machine, and when modified as commented, the formula becomes unrealizable.

slivingston:
Regarding the license, I just confirmed in private communication with the authors of Wring and Lily that the licenses have not been formalized, but they (the authors) are happy with the style of that of CUDD and we can get a formal statement if we need it. Moving forward, I recommend we assume Wring and Lily are under a just-in-time BSD 3-clause license.

PR #118, opened 2015-02-09
Offer "slugs" via tulip.synth and use --jsonOutput tulip.interfaces.slugs

slivingston:
In this pull request, the function `synthesize()` in the `tulip.interfaces.slugs` is updated to use the switch --jsonOutput for getting output in JSON, and `slugs` is now available via several functions in `tulip.synth`. [slugs](https://github.com/LTLMoP/slugs) is now regarded as an optional dependency. For the time being, meeting the dependency is somewhat complicated. While we can hope to evenually persuade upstream to re-merge [pull request #8](https://github.com/LTLMoP/slugs/pull/8) and thus make installation easier, "slugs" will remain a dificult dependency because upstream offers neither releases nor versions.

## Build and install "slugs"

At time of writing, the remote HEAD points at [cd9959e54b093560a85833550780121844150b0f](https://github.com/LTLMoP/slugs/commit/cd9959e54b093560a85833550780121844150b0f). The instructions below are appropriate for Ubuntu 14.04 with Linux x86_64.
```shell
git clone https://github.com/LTLMoP/slugs.git
cd slugs
```
Extract the [CUDD 2.5.0](http://vlsi.colorado.edu/~fabio/CUDD/) tarball into the `lib` directory. Direct link is <ftp://vlsi.colorado.edu/pub/cudd-2.5.0.tar.gz>; the SHA-1 hash of the tarball is 7d0d8b4b03f5c1819fe77a82f3b947421a72d629. Then,
```shell
cd lib/cudd-2.5.0
patch Makefile ../../tools/CuddMakefile.patch
make
```
At this step, CUDD is built in the manner requested by the [slugs instructions](https://github.com/LTLMoP/slugs/blob/cd9959e54b093560a85833550780121844150b0f/README.md). Now,
```
cd ../..
cd src
qmake Tool.pro 
make
```
At this step, there is an executable file named `slugs` in the current directory. Place it somewhere on your system path. E.g., a non-root way to achieve this is
```shell
mkdir -p ~/opt/bin
cp slugs ~/opt/bin     
export PATH=$PATH:~/opt/bin
```
Now we will install the Python package `slugs`, which was introduced upstream via [pull request #8](https://github.com/LTLMoP/slugs/pull/8),
```shell
cd ..
git checkout 23a75790547e058552f0a3f76415b3400b3145f4
tools/StructuredSlugsParser
python setup.py install
```
where above you may want to activate a Python virtual environment before the `setup.py install` command; otherwise, the last line may need a prefix of `sudo`.

Now test the result by going to TuLiP and, using the code as of this pull request,
```shell
./run_tests.py slugs
```

After merging into master, I plan to incorporate the above building instructions into the TuLiP User's Guide. My motivation to not do so yet is that we may be able to eventually provide an easier process of meeting the dependency.

### DIFF
--- a/tulip/interfaces/slugs.py
+++ b/tulip/interfaces/slugs.py
@@ -47,6 +47,9 @@ import slugs
 from tulip.spec import GRSpec, translate
 
 
+BDD_FILE = 'strategy_bdd.txt'
+
+
 def check_realizable(spec):
     """Decide realizability of specification.
 
@@ -63,12 +66,11 @@ def check_realizable(spec):
         fin.write(s)
     logger.info('\n\n structured slugs:\n\n {struct}'.format(
         struct=struct) + '\n\n slugs in:\n\n {s}\n'.format(s=s))
-    options = [fin.name, '--onlyRealizability']
-    realizable, out = _call_slugs(options)
+    realizable, out = _call_slugs(fin.name, synth=False)
     return realizable
 
 
-def synthesize(spec):
+def synthesize(spec, symbolic=False):
     """Return strategy satisfying the specification C{spec}.
 
     @type spec: L{GRSpec} or C{str} in structured slugs syntax.
@@ -84,8 +86,7 @@ def synthesize(spec):
         fin.write(s)
     logger.info('\n\n structured slugs:\n\n {struct}'.format(
         struct=struct) + '\n\n slugs in:\n\n {s}\n'.format(s=s))
-    options = [fin.name, '--jsonOutput']
-    realizable, out = _call_slugs(options)
+    realizable, out = _call_slugs(fin.name, synth=True, symbolic=symbolic)
     if not realizable:
         return None
     os.unlink(fin.name)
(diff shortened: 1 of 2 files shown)


necozay:
Just for the record: it is a bit of a pain to install `slugs` on Mac OS 10.8 and Ubuntu 12.04. It uses some fancy functions that only newer versions of gcc recognize. I tried to install it in the presence of Ruediger but was not successful. He seems to have some plans to make the installation easier though.

johnyf:
As of which commit of `slugs` was this difficult (`slugs` has no version number) ?

As of cd9959e54b093560a85833550780121844150b0f I did not find significant difficulties installing `slugs` on both Mac OS 10.8.5 with MacPorts 2.3.3 and Ubuntu 14.04.1 LTS. [This issue](https://github.com/LTLMoP/slugs/pull/4) may also be relevant.

necozay:
Just tried last week with the current head of master (cd9959e54b093560a85833550780121844150b0f). I have an old version of XCode (with `gcc` 4.2.1) and that was the main problem. I tried to get `gcc49` using MacPorts but I could find a port for `g++`.

PR #119, opened 2015-02-10
Improve style of bibliography

slivingston:
Consult the commit message of 3cd1351cea77344f03a7384b869f9865af34a4a4

I updated the existing citations in the User's Guide. Note that existing citation URLs in the source code (docstrings) are still valid. At time of opening, the changeset provided by this pull request only  touches files under `doc/` of the source tree.

To review,
```shell
cd doc
make html
```
then open `_build/html/bibliography.html` in a Web browser. Compare it with <http://tulip-control.sourceforge.net/doc/bibliography.html>.


johnyf:
As commented in that commit message, what about using `bibtex` and `sphinxcontrib-bibtex`, or `bibtexparser`, or `pybtex`?

slivingston:
The sophistication introduced by adding a dependency is not justified for our current usage. The four advantages that I imagine using one of the tools that you list (or some other such tool) provides are:

1. simplifying creation of a new reference, provided that the bibtex data for it is readily available;
2. storing a canonical form of the reference and thus being able to render it according to different styles automatically;
3. linking using a command like `cite{}`
4. easing the export of references from the documentation of TuLiP.

I think that items 1 and 2 are not well motivated because we have a relatively small number of references and it seems unlikely that the citation style will change often. As for item 3, the command would not be recognized in the source code docstrings, and thus manual linking to the bibliography of the User's Guide would still be necessary for citations in docstrings. Finally, I find item 4 to be motivating even for our present use-case, but the effort of an additional dependency imposed upon everyone who builds the docs does not seem justified by the reduced effort of those few who want to copy citations into bibtex format.

PR #121, opened 2015-03-21
Spec

johnyf:
Summary of changes:

- add `type` attribute to each AST node class, so that they can be distinguished inside `tulip`, and elsewhere. The reason is that a factory generates new classes each type that it is called. Therefore, just calling `spec.ast.make_nodes` twice suffices to create two identical families of classes, where `isinstance` fails when using a child from one family, with a parent from the other family. This can happen in `spec.translation`, as well as outside `tulip`.

  Using a `type` attribute makes all syntax algorithms agnostic of the origin of the AST classes. It implements the duck test (so the algorithms can be re-used for any AST that looks the same), and adheres to "explicit is better than implicit".

- add automatic split of an LTL string to a `GRSpec`, including checks that the given string conforms to the syntactic definition of GR(1) given by Piterman, Pnueli, Sa'ar
- check that primed system variables don't appear in environment safety
- check that initial conditions and recurrence don't contain primed variables (the check on recurrence shall be made optional when interfaced to a solver that supports recurrence of actions, as is the case with `slugs`)
- add ternary conditional operator to LTL syntax. The current syntax is `( ite expr, expr, expr )`, which is tentative and was selected as a compromise between uniformity (for n-ary operators) and context-insensitivity of the grammar.
- add truncation operator to LTL syntax. The syntax is `expr <<>> number`
- change LTL operator precedence and associativity, which were non-standard
- add `__repr__` to `GRSpec`, so that specs can be pickled.
- when generating parse table in `setup.py`, get the table file name defined in the parser module. So no duplication of constants (here the file name).
- various bugfixes and improvements to `spec`


johnyf:
The merge conflict is one line. I can resolve and merge after the pull request is reviewed.

johnyf:
For reference, some more details for why a `type` attribute was added (as discussed in an email):

1. This is a factory of classes from `tulip` that can be used anywhere (both inside and outside `tulip`.
  So it creates new class objects, and these are different from those instances of classes produced in `tulip.spec.ast`, and yet different from those in `tulip.spec.translation`.
  They do not share the same parent, only the class name of the parent is the same. For example:
  ```
  >> from tulip.spec import ast
  >> nodes = ast.make_nodes()
  >> u = nodes.Node()
  >> isinstance(u, ast.nodes.Unary)
  False
  ```
  This limitation led me to change approach.

2. Subclassing nodes precludes using `type(u).__name__`

3. I try to use `spec` as a "syntax" module, so that I don't have to rewrite certain functions just because of a slightly different AST. So I am using the attributes "operator", "value" as a minimal definition for compatibility between different sets of classes for syntax trees, facilitating code reuse (for expressions at least).

slivingston:
I will review this PR, but others may be curious about terminology. The various references to ducks, such as in [the message of the first commit on this branch](https://github.com/tulip-control/tulip-control/commit/d15b8a17a83b7e843c64ead8137881a2007f0068) are allusions to [duck-typing](https://docs.python.org/2.7/glossary.html#term-duck-typing).

slivingston:
Could you justify the creation of the ternary `ite` and truncation `<<>>` operators within the LTL syntax? In particular, are they not extensions that warrant the creation of a new specification language for Tulip?

johnyf:
Currently, I use them in code external to `tulip`. They are not used by any functions inside `tulip`. 9302a593fc4b1d65c918ec4d4bfdf1841c000185 updates the syntax in the user documentation.

slivingston:
Rephrasing my questions: I am not aware that those operators are regarded as LTL. As such, is it not the case that we should create a separate parser, etc., that extends the existing one, rather than assocating them with TuLiP LTL syntax?

I expect that we would want to do this eventually, e.g., if support for so-called signal temporal logic or other languages is added. Otherwise, we can drop "LTL" from "TuLiP LTL syntax" and instead have a specification input language that is the union of all languages of interest.

johnyf:
Indeed, the language is not propositional temporal logic, but (unquantified) first-order temporal logic (uQFOLTL). The existing parser parser parses a fragment of uQFOLTL, because the grammar includes nullary functions (integer and string variables) binary functions (+, -, *, /), and binary predicates (<=, <, ==, !=, >=, >). A separate parser could be created instead, but since the existing rules are not overridden, it is simpler to add two production rules.

slivingston:
Almost done reviewing. @johnyf could you verify that the test `test_split_gr1()` in `tests/gr1_test.py` as introduced in c9881dd51de38e014ba7f716f3c367abce49575e fails?

I expect that changing the test to compare lists instead of strings is sufficient to make it consistent with the return type of `tulip.spec.gr1_fragment.split_gr1()`. However, I hesitate because `split_gr1()` and `test_split_gr1()` were introduced in the same commit, c9881dd51de38e014ba7f716f3c367abce49575e, and have not changed since then, yet the tests are *failing* when I try them.

slivingston:
I studied the source of `split_gr1()` and am confident in my revised `test_split_gr1()`. You can ignore my above request to verify the failure.

Issue #105, opened 2014-09-12
sort out required and optional dependencies and separate test coverage accordingly


necozay:
`setup.py` makes a distinction between required and optional dependencies. So if one has all the required dependencies, the installation completes without an error. However, there is no such distinction in tests. After a successful installation, proceeding with tests typically results in a bunch of errors (mostly due to lack of some optional dependencies). This is quite confusing for new users. 

Maybe we should divide the tests into two: `-core` and `-all`, where `./run_tests.py -core` will run tests that can be run only with the required dependencies.

We might also want to consider revising what is required and what is optional (maybe no need to declare pure python dependencies `setup.py` can install automatically as optional).

Related issue: #48

slivingston:
A predicated test suite driven by `run_tests.py` is implemented as of cdfb3d6c62a3773aa9b99a16d8611bc615b1910a. Documentation appears in the README, the User's Guide, and `run_tests.py -h`.

In particular, note that `./run_tests.py` should always pass. It corresponds to the test family referred to as "base", i.e., `./run_tests.py` is equivalent to `./run_tests.py base`. It runs all tests involving only required dependencies, if any.

PR #122, opened 2015-03-22
Optional cvxopt

johnyf:
- create an ["extra"](https://pip.pypa.io/en/latest/reference/pip_install.html#examples) installation option called "hybrid" that can be installed with `pip install tulip[hybrid]`. The "hybrid" option installs `numpy`, `scipy`, `cvxopt`, and `polytope`.

- make `java` an optional dependency. Java is required by the `jtlv` solver bundled with `tulip`.
  As of 0a6893a69613f06ec50a910ba8884a7fc76b7682, the default solver has been decided to be `gr1c`, because `jtlv` is unmaintained and contains undocumented and/or erroneous behavior (Relevant `tulip-discuss` threads: [“the JTLV solver is dead”](https://sourceforge.net/p/tulip-control/mailman/message/33250491/) and [“`jtlv` outputs `#?#` depending on state numbering”](https://sourceforge.net/p/tulip-control/mailman/tulip-control-discuss/thread/540C44C4.1030806%40gmail.com/#msg32804942)).

- rm `unknown-commit` from version string, because it does not conform to PEP 440. As a result, `setuptools` fails to recognize such a version as valid, and parses it as `LegacyVersion`. When `pip` is instructed to fetch and install a tarball of a development version of `tulip` (or one of its forks), e.g., using `dependency_links` in `setuptools.setup`), it does not recognize the tarball contents as a valid `tulip` version.

- use only `install_requires`, and drop `requires` in call to` setuptools.setup`. In the future, the attribute `requires` [may be deprecated](https://wiki.python.org/moin/Distutils/Metadata), in favor of [`install_requires`](https://pythonhosted.org/setuptools/setuptools.html). Documentation on `requires` was found in [`distutils` docs](https://docs.python.org/2/distutils/setupscript.html#relationships-between-distributions-and-packages) and [PEP 314](https://www.python.org/dev/peps/pep-0314/). The packages `scipy`, `matplotlib`, and `networkx` use `install_requires`.

  My understanding is that `setuptools` considers both `requires` and `install_requires` and the latter is used to make a requirement more specific, in particular `setup_requires` vs `install_requires`. For this reason, it is preferred here (and to avoid duplication and inconsistencies between `install_requires` and `requires`).

- `nose` added to `tests_require`.

  Packages in `tests_require` are [only downloaded locally](https://pythonhosted.org/setuptools/setuptools.html#new-and-changed-setup-keywords), *not* installed. They are [installed only if `nose` is already present](http://nose.readthedocs.org/en/latest/api/commands.html), so that the command `python setup.py nosetests` be available. This partly defies the purpose of `tests_require`. If `nose` is added to `setup_requires`, then `nosetests` becomes available. Instead of all this confusion, it is easier to just add `nose` to `install_requires` (since it is a pure-Python dependency). For now it is placed in `tests_require`, because it is deemed optional.

  Packages from `extras_require` could be duplicated there. However, it may be simpler to add a `requirements.txt` and note that those are the development dependencies, and that tests require them. Relevant: #105.

- add `bugtrack_url` field to PyPI entry. This is a [PyPI-specific field](http://stackoverflow.com/questions/14459828/how-to-set-bug-tracker-url-in-setup-py-script) that appears frequently, e.g., [see this entry](https://pypi.python.org/pypi/kotti_software/0.1.4). Making it appear on PyPI informs people that GitHub is used, and several users may prefer to click and inspect the `tulip` repository.

Motivation:

1. do not place unnecessary constrains on users. If one wants to use `tulip.spec` (that has pure-Python dependencies) in a package,  the installation of that package should not fail because `cvxopt` fails to build. This prevents reuse of most of the code that exists in `tulip`.

  This use case applies to @johnyf, for code in `spec`. Currently, the respective `setup.py` first invokes `pip` to install a fork of `tulip` that installs easily. This can be done by using "dependency links" in `setuptools.setup`. However, these are deprecated. In the future, installation from outside PyPI will be even harder [PEP 438](https://www.python.org/dev/peps/pep-0438/). The change proposed here allows using the pure Python subset of `tulip` w/o installation failures.

  As a side-note, `setup.py` may still need to first call `pip` to install `ply` (`pip.main(['install', 'ply == 3.4'])`), so that it can build the parser tables. Nonetheless, this is a clean call to `pip`, and is equivalent to adding `ply` to `install_requires`. It is just a split of when each dependency is going to be installed, in particular shifting `ply` before the call to `setuptools.setup`, so that the tables can be generated to be included in the installation to `site-packages`.

  Few users that do not need `cvxopt` will make an effort to build `cvxopt`, and instead will avoid using `tulip` in their own work, and all packages that depend on `tulip`.

  Moreover, for those users that *do* need `cvxopt`, it is psychologically more encouraging to install the pure-Python part of `tulip` w/o trouble and start learning by using it, than to require them to wait (potentially a week or more), until they figure out how to build `cvxopt`.

2. If one calls a function that does require `cvxopt` or `polytope`, then an appropriate message can be printed.
  This approach is used by `networkx`, for example [here](https://github.com/networkx/networkx/blob/a822b491c6b53a2819915a8c620e97a36ba6c060/networkx/algorithms/centrality/current_flow_betweenness_subset.py#L102). The difference in `networkx` is that the `ImportError` is intercepted, and the message slightly refined. This is not a major difference from letting an `ImportError` be raised, without verbose `try...except` clauses in the source code. Noting in the readme that `ImportError` guards usage of code that requires optional dependencies should suffice (and can be cleanly reverted, if, in the future, the installation of `cvxopt` becomes robust. Otherwise, the `try...except` clauses would become redundant in case of such a reversal).

  The default `ImportError` that Python prints is already specific and informative enough. Any user with an elementary understanding of a Python installation will recognize that the package mentioned by an `ImportError` is most likely not present. Then, the user will confirm that this is the case, e.g., by invoking any of the following:
    - `python -c "import missing_package"`
    - `pip freeze | grep missing_package`
    - `pip show missing_package`.

  and install the package with `pip`, or re-install `tulip`, this time with the appropriate "extra" enabled.
  
  If the user neither recognizes the `ImportError`, nor searches to read its documentation or a discussion on the web, then teaching Python is beyond the scope of `tulip`.

3. An alternative is to separate `tulip` into smaller packages, based on common dependencies. Although this may be preferred in the future, for now it can be avoided by simply defining `extras_require`.

  Whether `tulip` would be represented better by multiple packages touches on issues of:

    1. software architecture, and
    2. neat and canonical structuring of theory and algorithms,

the discussion of which is beyond the scope of this issue.


slivingston:
Regarding the second item, while `gr1c` is now regarded as the default, I have not made all of the necessary changes for it to be so in the source code. E.g., at time of writing, the tip of master branch is 0a6893a69613f06ec50a910ba8884a7fc76b7682, and `examples/robot_planning/gr1.py` has

```python
mealy_controller = synth.synthesize('jtlv', specs)
```

I will make such changes and then merge them into the branch of this pull request before reviewing.

necozay:
I suggest ``install.rst`` under ``doc`` to be updated to reflect these changes before this is merged.

slivingston:
For conformance with PEP 440, specifically the section [Development releases](https://www.python.org/dev/peps/pep-0440/#developmental-releases), we need to include a number after "dev". E.g., compare with the approach of NumPy. On [line 110 of setup.py](https://github.com/numpy/numpy/blob/799a4c7e66b5e6ab6e9d48b29e386c39c991955e/setup.py#L110) in the [numpy repository](https://github.com/numpy/numpy), `.dev0+` followed by "Unknown" or a git revision identifier is used, which conforms to PEP 440. @johnyf Can you confirm this will succeed for your use-case? 

The following diff demonstrates how I will change `tulip/version.py`
```diff
@@ -59,6 +59,6 @@ if os.path.exists(path_to_hashfile):
             break
 
     if len(commit_hash) > 0:
-        version += ".dev." + commit_hash
+        version += ".dev0+" + commit_hash
 else:
-    version += ".dev"
+    version += ".dev0+Unknown"
```

johnyf:
I applied the above path to 8168668535962dcbe7d434ee136a5296702c3584 and installed successfully in a clean `virtualenv` with:

```
cd tulip-control
python setup.py sdist
pip install tulip==1.2.0.dev --find-links ./dist/
pip freeze
```

returns

```
cvxopt==1.1.7
decorator==3.4.2
networkx==1.9.1
numpy==1.9.2
ply==3.4
polytope==0.1.0
scipy==0.15.1
tulip==1.2.0.dev0+8168668535962dcbe7d434ee136a5296702c3584
```

slivingston:
## Main change from the OP

In commit da227485f990588886b0db7cc1e4b5bc2a4eb51b, I partially rolled back 8eb22889e39f3227e366745358fde13316581daa because I think that NumPy and SciPy are basic required dependencies that should not be regarded as "extra." One of the motivations of the OP was to have a subset of dependencies that allow the installation to proceed in pure Python. Obviously this is not true if NumPy is required. However, They are also part of the standard scientific Python stack (http://www.scipy.org/stackspec.html) and, as such, should have good availability through a variety of distribution media. The most common complaint that I have encountered is the difficulty of correct installation of CVXOPT, *not* NumPy or SciPy.

It is important to simplify usage of parts of Tulip that are of independent interest, like the subpackage `tulip.spec`. Doing so requires first addressing questions like what is core functionality of Tulip and whether Tulip should be a thin container of somewhat independent subpackages. But for this PR, it suffices to focus on making `cvxopt` optional and beginning to have more sophisticated capabilities with respect to requirements enforcement and satisfaction.

## Discussion

An important change brought about by this PR is that `pip` is now the preferred tool for local installations, i.e., the manual use-cases like fetching a tarball from SourceForge.net or using a copy of the repo. Thus, `pip install .` replaces `python setup.py install`. I note it here because this is a substantial (good) change that the original creator of this PR did not mention, and I could be missing something.

I have modified the [README](https://github.com/tulip-control/tulip-control/blob/8d1ceccd639c8b09a90fce0082f67e3dc6e84a10/README.rst) and the installation page in the [User's Guide](https://github.com/tulip-control/tulip-control/blob/8d1ceccd639c8b09a90fce0082f67e3dc6e84a10/doc/install.rst) accordingly.

One disadvantage of the approach to having extra requirements introduced by this PR is redundancy. If new extras are created in addition to "hybrid", then we may expect overlap in their dependencies. This will necessitate multiple occurrences of the same version constraint expression (e.g., `'numpy >= 1.7'`). These can be consolidated using a `dict`, but this is yet another complication in `setup.py`.

## Call for objections

Interested parties have had more than 1 month to object to the changes this PR will bring. Now that it is ready to merge, I will post a last call-for-objections on the mailing list with a deadline of 24 hours.

Issue #129, opened 2015-07-31
Allow partial match of inputs in reaction() of machines.py


iamkaushik5:
Hi, my name is Kaushik and I am a graduate student working with Necmiye Ozay on Tulip as part of a summer project and I wanted to bring this issue to everyone's attention regarding the reaction function in machines.py

Currently, the reaction function allows for stepping through a created mealy machine by providing the next input. However, the current function requires the input specification to match the entirety of the input spec in the mealy machine. However, when running most simulations, only a few inputs are required to identify a single path from the current state. Please see the code below:

  enabled_trans = [
            (i, j, d)
            for i, j, d in self.edges_iter([from_state], data=True)
            for k in inputs.iteritems() if k in project_dict(d, self.inputs).items()] #added
            # if project_dict(d, self.inputs) == inputs]

The commented line at the end is what is currently in the function. Notice how it requires that the input be directly equal to the input spec in the machine. Instead, the for statement above was added as a temporary fix. This for statement basically enables a transition if a single element of the inputs matches a single element of the output. This allowed me to simulate my example that I am working on because I only needed to provide the next state as input. However, this doesn't help for cases where multiple inputs are actually required to identify a single path from the current state. In that case, this fix would actually enable the same transiton twice, which will cause a determinism error because the program thinks there are two possible paths. Does anybody have any suggestions as to a permanent fix?

Second,
The error messages are not proper. The function seems to default to the 'this is not deterministic' error regardless of whether that is the issue. For example, if I sent inputs into the function that are not valid options from the current state, I get a determinism error, when really the error should say that 'these are not valid inputs'

Thanks.

-Kaushik Kannan

slivingston:
I may not understand your first item, but I think that requiring all inputs is correct and as intended. Your observation that knowing all inputs is not necessary in some cases is a particular case and, I can argue, not appropriate for this function. Perhaps you want to propose a new function, or some new way to request lazy evaluation of inputs when calling `reaction()`?

Regarding the second item, could you elaborate all of the errors and messages that you want?

Note that the error message is not as you quote it, but rather "must be input-deterministic". Input-determinism in different from determinism.

My comments above are based on the tip of master in the repo. Let me know if you are using a particular release of TuLiP.

necozay:
[...]

PR #130, opened 2015-08-28
Function for partial input matching, and more verbose error messages from reaction()

iamkaushik5:
This update addresses issue #129. the reaction function has been edited and a new function called reactionpart is for the partial match of inputs. 

The new error message for invalid inputs can be used in both functions. In the original function(reaction function), if you have an invalid input, the error message will display 5 possible valid inputs.

I have made 1 more function called reactionpart. This will allow for partial matching of inputs. For example, if only two inputs are required to identify 1 valid path from the current state, the function will allow you to only specify those 2 inputs. This function also includes the 'invalid input' error message and will display 5 possible inputs that match the keys of the user input. So, for example, if the user inputs "eloc: s5, level: 1" and if this this happens to be an invalid input, the exception will display 5 possible valid inputs that also include 'eloc' and 'level'.

### DIFF
--- a/tulip/transys/machines.py
+++ b/tulip/transys/machines.py
@@ -34,7 +34,9 @@ from __future__ import absolute_import
 import copy
 from pprint import pformat
 from random import choice
+import itertools
 from tulip.transys.labeled_graphs import LabeledDiGraph
+
 
 # inline imports:
 #
@@ -496,6 +498,7 @@ class MealyMachine(Transducer):
                 mask_func = masks[port_name]
                 self._transition_dot_mask[port_name] = mask_func
 
+
     def reaction(self, from_state, inputs):
         """Return next state and output, when reacting to given inputs.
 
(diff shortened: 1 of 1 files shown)


slivingston:
Comments here are regarding iamkaushik5/tulip-control-1@1048feb480cbfad60d6f809099689fd8ac2f20ac.

Feedback that must be addressed:

1. Indentation must be in blocks of 4 spaces. In particular, do not use the tab character for indentation.
2. There are two substantive and distinct changes. Please accordingly split the commit into two.
3. Use meaningful commit messages. "Update machines.py" is not really helpful because the fact that machines.py was modified is already available (e.g., via `git log --name-status`). E.g., in a commit introducing `reactionpart()`, you could write "Create reactionpart() to allow partial input matches".
3. In machines.py on lines 597-599, why is there code in the comments?
4. Could you justify placing the instructions in `reactionpart()` `#uncomment this to see the effect of`? I recommend the general design goal that reaction() and reactionpart() should present a reliable API, and their detailed internal operation need not be known. However, an exception could be that the variant obtained by uncommenting (as instructed) is worth considering later and is not obvious. However, when I read through the instructions and consider the resulting code, I think that you are merely trying to demonstrate your new error message code. Is that correct? If yes, it would be better to provide a separate example that demonstrates it, e.g., within the comment thread here or as a new or modified example in the repository.
5. The docstrings of `reaction()` and `reactionpart()` are identical. You should update the docstring of `reactionpart()` to indicate the partial evaluation.
6. In both `reaction()` and `reactionpart()`, you populate `trythese` with all possibilities but then only use the first 5. Why not break from the for-loop after 5 possibilities are found?

Style concerns:

* Delete trailing whitespace, as on lines 524, 536, 593
* In `reactionpart()`, we (or I) prefer the style where the docstring ends with `"""` on a line by itself.
* Try to avoid comments that state something immediately obvious, e.g., the one on line 579 in machines.py.
* Try to keep line lengths under 80.
* Multi-line comments should be on their own lines (not after code).

If you will not be able to address these in 2 days, then tell me and I can do it. Alternatively, if you address the crucial items, then I can handle the "style concerns" after merging.

slivingston:
Feedback addenda:

1. The exception message gives both inputs and outputs, despite describing them as "possible inputs". This can be addressed using `project_dict()`.
2. You modified the original implementation to use a for-loop instead of a list comprehension to construct `enabled_trans`. I expect your new code to be significantly slower than the original. To address this and my earlier comment about `trythese`, you could leave the list comprehension as in the original implementation and construct `trythese` by stepping through the first five items (at most) of `self.edges_iter()`.

slivingston:
Feedback addendum: 

Could you explain why computing permutations and combinations is necessary in `reactionpart()`?

As I understand it, the basic goal of `reactionpart()` is to allow an incomplete specification of input if there is precisely one enabled transition. Does it not suffice to modify the original implementation as follows?

```python
restricted_inputs = set(self.inputs).intersection(inputs.keys())
enabled_trans = [
    (i, j, d)
    for i, j, d in self.edges_iter([from_state], data=True)
    if project_dict(d, restricted_inputs) == inputs]
```

(I am done with the first reviewing.)

iamkaushik5:
Scott,

I don't think I will have any time to work on this now that my classes and
Grad Student Instructor duties have started. I want to thank you in advance
for taking care of these items for me. Just some comments on your feedback:

RE to item 4: this code can be deleted. It was either the original code or
some experimental code I was working with and I just wanted to leave it in
as a comparison, but it seems that the pull request takes care of showing
what the old code looked like.

RE to item 5: You are correct, it is there just to demonstrate the error
message.

RE to item 7: Yes we can break the for loop after 5 possibilities, but I
guess I left it more open in case we wanted to add a way to view all
possible inputs.

RE to Addenda 1, item 1: In what way do you mean that it gives both inputs
and outputs? The only thing I am collecting are possible inputs. I only
ever append trythese with project_dict(d, self.inputs), which is only the
inputs and I display trythese with the error message.

RE to Addenda 2 item 1: You are correct. The basic goal is to allow for
partial inputs, if those partial inputs are enough to identify one possible
enabled transition. For example lets say we are in a particular state that
has 3: possible paths. One is identified by the input: ('eloc' : s0,
'level': 1, 'park': 1, 'home': 0) another is identified by the input
('eloc': s1, 'level': 0, 'park': 0, 'home': 1) and another by ('eloc': s1,
'level': 1, 'park': 0, 'home': 1) . Now clearly one of these paths can by
identified as long as we know the value of eloc and level, so I would like
the user to be able to just input those 2 items and be ok. The reason I did
permutations and combinations is 1, I don't want order to matter. If the
user inputs eloc first and level 2nd or level first and eloc 2nd, it
shouldn't matter. 2, I want to be able to find all possible input
combinations of ONLY level and eloc so that I can display them in my error
message. If the user input 'eloc': s3, 'level': 1 is invalid, I want the
error message to say here are 5 valid inputs of eloc and level. As opposed
to saying here are 5 full length inputs.

Now I'm not sure what your intersection command that you listed above does
exactly, but I hope that my explanation above will allow you to see what
I'm getting at and why I used permutations and combinations. Your code is
very concise and if your code does account for all of the above then I will
wish I had known about that command earlier and blame my inexperience with
Python!

Also, I want to add that in the case that the user's partial input is not
enough to identify 1 path, the existing error message will catch it because
the function will essentially populate 'enabled transitions' with more than
1 item, and that will error as usual.

Thanks for all your help and best of luck with the rest of the Tulip
development. If you have any other clarification questions, I will do my
best to answer them.

P.S. If there are any further action items on this or any other area of the
tulip dev that you need from my end, I can take note of them and pass them
on/explain them to the next student that will be helping Necmiye with this
project(who is currently TBD).

Sincerely,
Kaushik Kannan

slivingston:
I am just going to close this PR and commit directly to master branch two changesets that provide the desired features. I have already confirmed my implementation [sketched above](https://github.com/tulip-control/tulip-control/pull/130#issuecomment-138426643). Please direct additional discussion to #129.

Regarding permutations and combinations, note that the `dict` type is not ordered. Unless I am missing something, casting to an ordered type and then enumerating permutations is not necessary and will result in much worse performance.

slivingston:
Replying to a question "In what way do you mean that it gives both inputs and outputs?" from [above](https://github.com/tulip-control/tulip-control/pull/130#issuecomment-138672132), check [line 529 in your modified machines.py](https://github.com/iamkaushik5/tulip-control-1/blob/1048feb480cbfad60d6f809099689fd8ac2f20ac/tulip/transys/machines.py#L529). In that line of `reaction()` , `d` is added to `trythese`, where `d` is the entire edge data dictionary, which includes both inputs and outputs. Apparently the same error does not occur in your proposed `reactionpart()`.

PR #134, opened 2015-11-11
interface to gr1py, a new tool for GR(1)

slivingston:
https://github.com/slivingston/gr1py
https://pypi.python.org/pypi/gr1py

"gr1py is an enumerative (or concrete) reactive synthesis tool for the GR(1) fragment of LTL. It is pure Python."

I recommend reviewing PR #133 first, because this is based on it. To run tests, try

    ./run_tests.py -f gr1py

Notes with comments:

* There is no reference to `gr1py` in the User's Guide nor among the checks of optional dependencies in `setup.py`. I recommend leaving it this way until we have experimented more with it, and possibly until the interface becomes more sophisticated.
* There is currently no declared preference about which versions of `gr1py` are supported. We should wait until the next release to decide it because `gr1py` could continue to progress between now and then. However, if you want to declare something now, use at least 0.1.0.
* A quick demonstration of the new interface is to modify `examples/robot_planning/continuous.py` to have `ctrl = synth.synthesize('gr1py', specs,` on line 101.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ numpy==1.7
 scipy
 
 # easy extras
+gr1py>=0.1.0
 pyparsing==2.0.3
 http://pydot.googlecode.com/files/pydot-1.0.28.tar.gz
 
--- a/tests/gr1pyint_test.py
+++ b/tests/gr1pyint_test.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+"""
+Tests for the interface with gr1py.
+"""
+import logging
+logging.basicConfig(level=logging.DEBUG)
+logging.getLogger('tulip.spec.lexyacc').setLevel(logging.WARNING)
+from nose.tools import raises
+import os
+from tulip.spec import GRSpec, translate
+from tulip.interfaces import gr1py
+
+
+class basic_test:
+    def setUp(self):
+        self.f_un = GRSpec(
+            env_vars="x", sys_vars="y",
+            env_init="x", env_prog="x",
+            sys_init="y", sys_safety=["y -> X(!y)", "!y -> X(y)"],
+            sys_prog="y && x")
+        self.dcounter = GRSpec(sys_vars={"y": (0, 5)}, sys_init=["y=0"],
+                               sys_prog=["y=0", "y=5"])
+
+    def tearDown(self):
+        self.f_un = None
+        self.dcounter = None
+
+    def test_check_realizable(self):
+        assert not gr1py.check_realizable(self.f_un,
+                                          init_option="ALL_ENV_EXIST_SYS_INIT")
+        self.f_un.sys_safety = []
+        assert gr1py.check_realizable(self.f_un,
+                                      init_option="ALL_ENV_EXIST_SYS_INIT")
+        assert gr1py.check_realizable(self.f_un,
+                                      init_option="ALL_INIT")
+
+        assert gr1py.check_realizable(self.dcounter,
+                                      init_option="ALL_ENV_EXIST_SYS_INIT")
+        self.dcounter.sys_init = []
+        assert gr1py.check_realizable(self.dcounter,
+                                      init_option="ALL_INIT")
+
+    def test_synthesize(self):
+        self.f_un.sys_safety = []  # Make it realizable
+        g = gr1py.synthesize(self.f_un,
+                             init_option="ALL_ENV_EXIST_SYS_INIT")
+        assert g is not None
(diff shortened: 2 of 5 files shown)


necozay:
In some TuLiP examples (including those that involve abstractions), we create explicit models in the form of transition systems and then encode the transition system in temporal logic to interface symbolic solvers. With `gr1py` (as of this pull-request), it looks a bit redundant to go from TS to spec and back to an explicit TS compatible with `gr1py`. Do you envision this redundancy to be eliminated to some extent (I can imagine this might be tricky in some cases) when you say "interface becoming more sophisticated", or maybe compatible transition system structures to be used in TuLiP and `gr1py`?

johnyf:
There are a few comments worth mentioning here.

An informal observation (I do not claim to have formally proved it, but it seems simple). In problems arising from abstractions, enumerated graphs are constructed. A formula that encodes such a graph has size at least as large as the graph, because it mentions every node, and every edge. This is special to the problems in `tulip`, and does not apply to problems that arise from "unfolding" a program. As a result, the resulting specification is about-linear in the state space (assuming the state space contains only states resulting from discretization).

Due to the previous size comparison, the cubic complexity of GR(1) synthesis can be misleading for formulae resulting from discretization. With thousands of states, a formula with thousands of disjuncts results. However, this doesn't lead to a blowup in synthesis, because the formula is already linear in the state space, so its cubed length saturates in the complexity analysis by the number of states.

This argument suggests that it may not be too bad to run the fixpoint algorithm on an enumerated representation, instead of a symbolic one.

Another note is the comment in 2012 Bloem et al. JCSS that with an enumerated data structure the complexity is better.

About the [issue above](https://github.com/tulip-control/tulip-control/pull/134#issuecomment-156609870) of converting twice:

1. enumerated graph in `tulip` -> formula
2. formula in `tulip` -> enumerated graph in `gr1py`

I would assume that this is not a performance issue, unless proved otherwise with profiling.
One reason is the enumeration mentioned above, which implies that the enumerated graph is about as large as the formula, so the amount of computation for conversions is linear in that size.

Another reason for converting everything to formulae, and then letting `gr1py` enumerate the state space is to unify handling of enumerated graphs and (string) formulae given as input.
This ensures a clean, standardized interface between the frontend (abstraction etc.) and the backend(s) (`gr1py`, `gr1c`, etc.).

For example, the graph representations inside `tulip` can change completely, the internals of `gr1py`, but as long as the frontend outputs the same syntax that the backend understands, the implementors of the frontend and backend do not need to know anything about the internals of the other part, or modify their code.

In a situation where I am parsing-flattening repeatedly, and was once doing that with manipulation of abstract syntax trees as enumerated graphs, the former approach has proved much cleaner and modular, whereas the latter is error prone.

A side comment is that a symbolic solver in Python has been interfaced to within the branch `python_symbolic_streett_1_interface` and will be candidate for merging after this one. So, we will have both types of solvers, and can also compare. I suspect that this may lead to observations similar to the comparison between enumerative and symbolic model checkers for hardware vs software systems (due to structure of the associated BDDs).

necozay:
I agree with the modularity comment (performance to be determined).

BTW, we have reported a similar observation in the switching protocol synthesis paper with Fei Sun et al. (ACC 14) that implements some fixed point algorithms on explicit graph representation. 

For abstractions, the transition graph is sparse (i.e., the branching factor is low). For instance, if you are abstracting continuous-time systems, transitions are limited to neighbors (and discrete-time systems modeling physics do not jump too further in one step for reasonable sampling frequencies). This makes certain explicit data structures for system representation advantageous at least from a practical standpoint. Similarly for the controller representation. 

Would be interesting to see if our observations carry over to these new set of solvers.

johnyf:
For the record, as of 19a17b15e358845900e193ccae9e659170a23ebe, on:

- Debian GNU/Linux 8 (jessie) 32-bit
- Intel® Core™2 CPU 6700 @ 2.66GHz × 2
- 2.0 GiB
- Python 2.7.8
- `networkx` 1.10

with `gr1py`:

```shell
$ /usr/bin/time -f "%E real, %U user, %S kernel, %P cpu %M max (KB)" python gr1.py
0:32.37 real, 30.45 user, 0.37 kernel, 95% cpu 35508 max (KB)
```

and with `gr1c`:

```shell
$ /usr/bin/time -f "%E real, %U user, %S kernel, %P cpu %M max (KB)" python gr1.py
0:02.06 real, 1.82 user, 0.09 kernel, 93% cpu 42808 max
```

Note that `gr1.py` uses a [one-hot encoding](https://en.wikipedia.org/wiki/One-hot).

Issue #115, opened 2014-12-16
Add STL syntax to TuLiP (specs module?)


murrayrm:
It would be very useful to be able to start using STL in TuLiP specifications, including (eventually) the quantitative semantics, which will allow for things like robust plans.  Depending on how many changes to the specs module are required, this could be a 2.0 type of feature.  But if we can start thinking about ways to add this in, I think it would be quite useful in the long run.

slivingston:
related work: <https://github.com/mvcisback/py-stl>, "Python library for working with STL syntax".

slivingston:
nontrivial application of py-stl is <https://github.com/mvcisback/py-blustl>

murrayrm:
Good find!  We should definitely have a look at this and see if we can find a way to use/support `py-stl` as a way of getting STL specs into TuLiP.  The code looks very early (no documentation yet), but being done by a grad student in Sanjit Seshia's group.

vraman:
py-blustl is under active development, but not far from being ready for
external use. I am in touch with the author, so can relay TuLiP-specific
feature/documentation requests if needed.

Vasu

slivingston:
update: https://github.com/mvcisback/py-blustl now redirects to https://github.com/mvcisback/magnumSTL
in the root of the sourcetree there is a [nontrivial IPython Notebook](https://github.com/mvcisback/magnumSTL/blob/63e1132544706d5bbd7590735bb7b6f12d6a661c/Untitled.ipynb)

slivingston:
@vraman this task was originally assigned to you, but if you do not have any prior work waiting to be committed, can I do it?

I am working on a project for which having STL in TuLiP would be useful.

murrayrm:
Is anyone working on this functionality?  I have some undergrads interested in working on projects next quarter and could get someone started in this direction.

vraman:
In an in-person conversation with Scott last month, I gave him my approval
to take over this feature.

Vasu

slivingston:
@murrayrm I am working on it, motivated in part by a project that might use it. I hope to move quickly on it in the coming month.

PR #135, opened 2015-11-16
interface to symbolic Python solver for GR(1) games

johnyf:
The solver is in [`omega.games.gr1`](https://github.com/johnyf/omega/blob/82752680d83d21c14c1f04af471f0e75d59dc99c/omega/games/gr1.py). Please install branch `master`, because `omega` 0.0.3 on PyPI does not work with [`dd >= 0.2.0`](https://github.com/johnyf/dd) (latest on PyPI).

The interface added to `tulip.synth` uses the Streett(1) solver of `omega.games.gr1`.
A Rabin(1) solver (= GR(1) counter-strategies) is also available and tested in `omega.games.gr1`.
It is not interfaced at this time, because `synth` does not return counter-strategies, but the addition of counter-strategies would be a small extension of `tulip.interfaces.omega`.

The function `tulip.interfaces.omega.is_circular` is added, to check whether the system can "trivially" win a game. Roughly, this means that, from some initial states, the system has a winning strategy for at least one persistence goal (= negated environment recurrence, as usually called in GR(1) games).
The `is_circular` function should allow easily diagnosing specifications with "circular" dependency of environment and system liveness goals.

The package `omega` and its dependencies are pure Python. This ensures portability, and for most relatively small problems, it should work fine. For large problems, `dd` interfaces to CUDD via Cython. With a try-except import in the module `omega.games.gr1` (to be added in the next release), it can use CUDD whenever present.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ numpy==1.7
 scipy
 
 # easy extras
+omega>=0.0.4
 pyparsing==2.0.3
 http://pydot.googlecode.com/files/pydot-1.0.28.tar.gz
 
--- a/tests/omega_interface_test.py
+++ b/tests/omega_interface_test.py
@@ -1,0 +1,108 @@
+"""Tests for interface to `omega.games.gr1`."""
+import logging
+from tulip.spec import form
+from tulip.interfaces import omega as omega_int
+from tulip import synth
+
+
+logging.getLogger('astutils').setLevel('ERROR')
+logging.getLogger('omega.logic').setLevel('ERROR')
+
+
+def test_grspec_to_automaton():
+    sp = grspec_0()
+    a = omega_int._grspec_to_automaton(sp)
+    dvars = dict(x=dict(type='bool', owner='env', dom=None),
+                 y=dict(type='bool', owner='sys', dom=None))
+    assert a.vars == dvars, (a.vars, dvars)
+    r = a.action['env']
+    assert r == list(), r
+    r = a.action['sys']
+    assert r == ["x' -> y'"], r
+    r = a.win['<>[]']
+    assert r == ['!(!x)'], r
+    r = a.win['[]<>']
+    assert r == ['!y'], r
+
+
+def test_synthesis_bool():
+    sp = grspec_0()
+    h = omega_int.synthesize_enumerated_streett(sp)
+    g = synth.strategy2mealy(h, sp)
+    # fname = 'mealy.pdf'
+    # g.save(fname)
+    # pd = nx.to_pydot(g)
+    # pd.write_pdf(fname)
+    assert g is not None
+    assert len(g.inputs) == 1, g.inputs
+    assert 'x' in g.inputs, g.inputs
+    dom = g.inputs['x']
+    dom_ = {False, True}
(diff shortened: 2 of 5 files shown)


johnyf:
[`omega` v0.0.4](https://github.com/johnyf/omega/tree/v0.0.4) was released, so that it be compatible with `dd >= 0.2.0` on PyPI. `omega` v0.0.3 allowed `dd == 0.2.0` as a dependency, which was incompatible.

Avoiding this touches on [semantic versioning](http://semver.org/):

"Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable."

In other words, this is still compatible with semantic versioning. In any case, while in `0.y.z` I try to use `y` for changes that break backward compatibility.

slivingston:
Tested using omega repository at [0c687c804f8374eeb38344992745446b46ae24b5](https://github.com/johnyf/omega/commit/0c687c804f8374eeb38344992745446b46ae24b5).

Discussion:

* tulip/interfaces/omega.py does not include the usual copyright and license declaration at top of the file. In private communication with me, @johnyf observed that there is a license file at the top level as of 36f3d4135023cf633871e0ff32f896e4a4198ad9 and asks whether including a copy of the BSD license text is necessary in every file. We should discuss it outside this pull request, e.g., on the mailing list.

* I enjoy your choice of `synthesize_enumerated_streett()` instead of `synthesize()`, e.g., which is used in interfaces/gr1c.py. What do you think about introducing `synthesize_enumerated_streett()` as an alias for `synthesize()` in the other interfaces and declare `synthesize()` as deprecated (or renamed)? One motivation is that several of these tools can synthesize for more than one type of game.

* _grspec_to_automaton() in tulip/interfaces/omega.py could be of sufficiently general interest to export.

* After merging, I want to try this new solver on more benchmarks. The times listed below motivate further consideration of it as the new required GR(1) game solver, instead of gr1c.

For comparison with times provided in a [comment on the pull request that introduced gr1py](https://github.com/tulip-control/tulip-control/pull/134#issuecomment-156911329), as of 43c6881b02b51dbd6ec0b38a60c07033c4181884, on
* Ubuntu 14.04.3 LTS ("trusty"), x86_64 GNU/Linux
* Intel Core i7-4700MQ CPU @ 2.40GHz
* 8 GB RAM
* Python 2.7.6
* networkx 1.9.1

with `gr1c`:
```sh
/usr/bin/time -f "%E real, %U user, %S kernel, %P cpu %M max (KB)" python gr1.py
0:00.89 real, 0.42 user, 0.05 kernel, 53% cpu 32880 max (KB)
```

with `omega`:
```sh
/usr/bin/time -f "%E real, %U user, %S kernel, %P cpu %M max (KB)" python gr1.py
0:01.96 real, 0.66 user, 0.11 kernel, 39% cpu 35708 max (KB)
```

slivingston:
Regarding my motivation for using `synthesize_enumerated_streett()` in the other interfaces, gr1c includes `gr1c rg` for solving reachability games. A [brief description](http://slivingston.github.io/gr1c/md_spc_format.html#reachgames) is given on the manual page about specification formats. There is also Lily. An upcoming one is STL; cf. #115

PR #136, opened 2015-12-05
Um

necozay:


slivingston:
Was this pull request accidentally created?

necozay:
No, it was not accidentally created. And it is already merged to the `aug_sys_um` branch. Then, due to the fact that people at the Hackathon could not push to `tulip-control:aug_sys_um` branch, we decided to continue the development on `necozay:aug_sys_um` branch, which will be eventually merged to `tulip-aug_sys_um` when it is ready.

PR #144, opened 2016-03-29
refactor setup

johnyf:
The `setup.py` file is rewritten (step-by-step) to modernize it, drop old checks (`jtlv`, Java, `cvxopt` is optional in branch `master` of `polytope`, other solvers besides `gr1c` have been interfaced to), and use infrastructure available in `setuptools` and `pip` (`install_requires`, `extras_require`, `requirements.txt`).

 A duck approach is taken, letting some optional dependencies fail when needed (`matplotlib`), instead of warning (warnings printed during installation are less likely to be read than a note in the README). For solver interfaces, the checks are performed locally, e.g., by the interface to `gr1c`, to `omega`, etc.


slivingston:
The comment in the beginning of the definition of package_jtlv() is too general. Shall I delete it, or make it specific to the JTLV-based solver ?

johnyf:
I think that it is better to delete it.

slivingston:
59d9f365d7cc2dd1511be8d71ea58b80ee54a4b3 is not relevant to this PR, but it is motivated to avoid spurious failures on Travis CI.

slivingston:
I added the appropriate SHA-256 hash to the changeset [referred to above](https://github.com/tulip-control/tulip-control/pull/144#issuecomment-204831392), resulting in a0a7186403e82c282ae6db11b30e6898cee6f128.

slivingston:
@johnyf: these are good improvements. I agree with the design principle asserted in 1ad10295d780a44d1bdda0f8e35f10f1893e4b40, and it is worth repeating here in case discussion is desired:

> Each module in `tulip.interfaces` is responsible for checking the presence of the tool it communicates with.

slivingston:
What is the significance of the comment "setup" introduced in da491e3ac406d0478cd14b610546ed3b43de923a ?

Otherwise, this is ready to merge.

johnyf:
That comment is not very significant and can be deleted. Its main purpose was to visually delimit the version definition from augmenting package data and calling `setup`, though the former is related to setup as much (and more) as the latter, thus the comment can be misleading.

slivingston:
This was merged, but I accidentally deleted the remote branch before the merge occurred, so GitHub incorrectly marks it as "closed" not "merged". Restoring and deleting again the branch does not change this behavior of GitHub.

slivingston:
@johnyf If you want, I am happy to push a merge commit now to emphasize that the merge occurred and is a result of this pull request.

johnyf:
I am fine with the existing history, which is linear.

PR #145, opened 2016-03-29
Version in setup and use gitpython

johnyf:
The description of 114e010 gives a short summary. This adds to #144.

Old approach
---------------------

Does not assert that the version number in the tag
matches the version defined in the code (`version.py`).
So, a commit tagged `v0.0.0` can be released as
`v1.1.1` if that value is in `version.py`.

The old approach is nested and not as simple,
because it first dumps the commit hash to a file,
and then loads it back, together with the version number,
by importing `tulip/version.py`.

This requires importing `version.py` without importing `tulip`,
so a call to `imp.load_module`.

The version number is defined inside `version.py`,
instead of `setup.py` (which is at the top).

Dumping the hash and loading it back again is not as simple as just dumping information.
It does work in case that a development version is packaged and distributed
(because the hash is shipped with the `commit_hash.txt`),
but that is a corner case that I've never done myself.

Arguably, it is a corner case that should be discouraged, instead of supported,
because it would imply that one develops by passing back and forth tarballs,
instead of cloning the repository.

Some code is replaced by calls to `gitpython`,
which introduces a development dependency.
It is not clear whether this is an improvement.
But I think that using `gitpython` makes the code more readable.

The case of a tag that starts with `'tulip-'` need not be handled in version `1.2.0`,
because no future release will be tagged with that prefix.

New approach
-----------------------

Duck failure: just ignore that `gitpython` is not installed,
or `.git` absent, or `git` not in the path, or something else wrong.
A developer is supposed to have installed development dependencies like `gitpython`.

The only use case that I can see a real difference in is when
uploading a release with `gitpython` absent and having forgotten to tag.

With the old approach, this would result in either a local version identifier (containing a commit hash),
which cannot be pushed to PyPI, or an "unknown" local version.
This protects against releasing without tagging and having `git` etc.

The new approach does not guard against this.
A guard cannot be introduced, because a release should install itself as normal,
without marking itself as "unknown" because it didn't find `git` information.

A developer is supposed to follow a protocol for releasing.
After all, there are so many things that can go wrong,
but the code doesn't guard against every possible omission.

If the working directory is dirty,
then this is noted to the local version specifier.
This is useful during experiments,
and improves the old approach.

The proposed code is being used in the packages `dd, omega, openpromela`,
except for raising an `AssertionError` when tag information is found that does not match the version in code (should be fixed in those packages).
The `try-except` within `run_setup` is motivated by
the complexity in how many ways attempting to retrieve
a commit hash and tag can fail.

The version module is hidden (`_version.py`), because it serves an auxiliary purpose.


slivingston:
@johnyf Do you have any preference about whether this is included in the next release, which is version 1.2.1, or the release after it?

johnyf:
I have a preference to include these changes in the next release, so that installation be more pythonic and use existing facilities (like `pip`) where possible.

slivingston:
For posterity, I understand references to `gitpython` to be to GitPython as [on PyPI](https://pypi.python.org/pypi/GitPython/) and [on GitHub](https://github.com/gitpython-developers/GitPython).

johnyf:
Yes, to install `pip install gitpython`, but to use `import git`.

slivingston:
@johnyf As part of my review, I introduced [conda](http://conda.pydata.org/docs/index.html) (also consult Anaconda documentation at <https://docs.continuum.io/anaconda/index#conda-anaconda-and-miniconda>) as the means by which we install numpy and scipy on Travis CI. The change is intended to reduce the time required to install dependencies before testing.

johnyf:
I am working on replacing `tests/version_test.py`. This will be a test for `setup.py` itself. Little information seems to be available for practices about testing `setup.py` (which is absent from an installation), so I will follow a simple approach of `try`-`import`.

johnyf:
Developers need to delete `commit_hash.txt` and `version.pyc` that are ignored by `git`. Perhaps we could add a `Makefile` with a `clean` rule. The `Makefile` would be excluded from the Python distribution archive, so it shouldn't cause much confusion.

On the other hand, presence of a `Makefile` in the top directory may suggest that `tulip` involves installation mechanisms other than what is standard for Python packages. Let me know whether you think we should have a `Makefile` for developer convenience. I find it convenient to have a [`Makefile`](https://github.com/johnyf/dd/blob/94bc402664fec5ab2f2c76459db598d385b996a7/Makefile#L49) in `dd`, but `dd` does include Cython, so the `Makefile` is not entirely misleading (even though it's disjoint from the installation mechanisms).

slivingston:
If you are referring to the vestigial commit_hash.txt and tulip/version.pyc that will remain after this branch is merged, then I think that it is enough to tell everyone to manually delete them.

slivingston:
I noticed the failure on Travis CI because `mock` is not found, despite it [being pre-installed](https://docs.travis-ci.com/user/languages/python/#Pre-installed-packages). The conda installation is hiding `nose` and `mock`. I will try to recover them.

slivingston:
Locally I have found that your new version_test.py passes.

johnyf:
`mock` has been introduced as a dependency for tests. The only reasonable alternatives I could think of are:

1. mock things myself: started doing so, but I need to mock the `git` package imported inside `setup.git_version`. The `mock` package provides a ready-made, widely used solution.

2. use `gitpython` with a *real* `git` repository created within the `tests/` directory. I consider this approach dangerous, because:
    - if the test fails, then it leaves a `.git` repo behind. This can have various effects, possibly including failure of the test if re-run, because of inability to overwrite an existing repo (assuming some commit is made).
    - if tests succeed, still the dummy repo needs to be deleted. This may sound innocuous, but calling manually the same test script from the wrong directory (say from within the top directory) could possibly erase our own work on `tulip` itself.
    - touching files in an automated way nested within tests is dangerous
    - involving `gitpython` functionality in the test reduces its credibility. In other words, what are we testing? `gitpython` against `gitpython`? No.

For the above reasons, I decided to use `mock`. Another reason was that I haven't used `mock` before and wanted to try it.

johnyf:
Unfortunately, `gitpython` is used during testing, even though it has been almost avoided. The reason is that I [needed to mock `git.GitCommandError`](https://github.com/tulip-control/tulip-control/blob/110adcb95051b79f343b0c98d7807c58b6529ad7/setup.py#L63) (because one return path from `git_version` relies on recognizing this exception), in a way that it is both mocked *and* an `Exception`. I tried, but couldn't find a way to do so, not even by subclassing both the mocked class and `Exception` together.

So, I ended up [using `git.GitCommandError` itself](https://github.com/tulip-control/tulip-control/blob/110adcb95051b79f343b0c98d7807c58b6529ad7/tests/version_test.py#L43), and mocking the rest.

johnyf:
Compared to the previous version management mechanism, one lost capability is distributing a development version tarball that `pip` recognizes as a development version, but which installs itself as a release. In other words, a tarball that used a local identifier during packaging (for example `tulip-1.2.1.dev0+f4adba5c320e0e3d1da50eb248ee5793606cbc8e.tar.gz`), but which installs itself as a release (`tulip-1.2.1` in this case), because at installation time it didn't find any `git` repository.

This use case has arisen in practice for me, but I would argue that it is due more to bad practice (imposed by uncontrolled factors), rather than something to worry about. What this indicates is that:

1. dirty workspaces should *not* be used for serious benchmarking and testing,
2. development versions should be installed from a repository.

The corner case of supporting the tarball-based distribution of a development version raises the question of whether this use case justifies the associated complexity.

Another shortcoming is that the local version identifier is fixed to be the commit hash. But that was so before too, and I don't think that the tag management needs to happen within the python version identifier. Tags can remain within `git`. This also ensures that the version identifier has a unique form (which was also the case previously, and is arguably canonical).

slivingston:
For each test, the fully qualified function name is replaced by its docstring, if one is given, in the results printed by `nose`. Knowing the function name is useful because the source code for it is easily found. I recommend one of the following changes:

1. the docstrings of test_git_version() and assert_pep440() are changed to comments (and thus, not found by `nose`); or
2. the docstrings have prefix with the module name and function name.

The second option introduces redundant information that would need to be manually updated if the function or module name changes.

johnyf:
I noticed this too. However, I did write the docstrings, because static analysis was complaining, and I think that it is right. It is the behavior of `nose` that causes this issue. There is plugin to disable this behavior [nose-ignore-docstring](https://pypi.python.org/pypi/nose-ignore-docstring). I tested it, and it works.

I agree that No.2 should be avoided. If `nose-ignore-docstring` is an acceptable solution, then we can avoid No.1.

A relevant issue: https://github.com/nose-devs/nose/issues/294 that mentions another `nose` plugin, called [`disabledoc`](https://pypi.python.org/pypi/disabledoc).

slivingston:
Regarding [your comment above](https://github.com/tulip-control/tulip-control/pull/145#issuecomment-226376901) about having to use `gitpython` during testing, I recommend that `'version_test'` is moved out of the base family of tests defined in run_tests.py and into the "full" family. Since we currently only perform the base family of tests on Travis CI, the command `./run_tests.py -f version_test` can be added.

johnyf:
This sounds reasonable, until a way is found to mock the typed exception (types are harmful).

johnyf:
On the other hand, since the tests require `nose`, anyone running them should also install `mock` and `gitpython`, both of which are as easy to install as `nose` (`git` shouldn't be needed, because the repo is mocked).

slivingston:
Is there a use-case that involves accessing the docstrings of tests? I understand that using comments instead of docstrings precludes introspection, but will anyone want to automatically extract `__doc__` from those functions? If a test fails, then my first action is to read the source code of the test. If the test fails for a user, she might copy-and-paste terminal output and send it to us.

slivingston:
As a design principle, users who are not hacking on TuLiP itself should be able to run tests with as few as possible additional dependencies.

slivingston:
I agree that it is reasonable to expect users who can get `nose` to be able to get `mock` and `gitpython`. However, they might not use those packages again after completing (and passing) tests of TuLiP. Thus, requiring the packages contributes to installation creep on their machines.

slivingston:
I do not think that this is a critical issue, so I am OK to simply require `mock` and `gitpython` for the base family of tests. As we explore future applications and get user feedback, we might decide later to modify when these are required.

johnyf:
Looking at the code of these plugins and `nose` docs, it's a matter of 5 lines, so I added them to `run_tests.py` in e9d8684.

I agree that it is convenient, for myself too, to run `nose` without plugins or arguments. I don't know a use case for extracting docstrings from tests.

Noting yet another plugin with the same purpose, [nosenodocstrings](https://github.com/MarechJ/nosenodocstrings).

Also, we may want to add a `requirements_dev.txt` that contains `nose, mock, gitpython`, because `requirements.txt` pins packages other than these, and `tests_require` includes also `matplotlib`, which isn't easy to install. I'm not in favor of adding an extra file just for this purpose though, only mentioning it as a possibility. The test dependencies have already been noted at 2 different places, and when they fail, it is very clear from the error what the user should install.

slivingston:
I agree with your comment in the commit message of 1b7a82aae3ddf844c8a3718fed882fee7e74866c that it is just the `nose` API, so no additional attribution is needed.

slivingston:
These are good changes and make the installation and version labeling more robust. The particular case of dev tarballs is not handled well, [as you observe above](https://github.com/tulip-control/tulip-control/pull/145#issuecomment-226378504) and [in the OP](https://github.com/tulip-control/tulip-control/pull/145#issue-144193059), but using them is not good practice.

slivingston:
As I have begun to work more with the changes introduced in this pull request, I am finding that installing Tulip is now a fragile process that often fails for me and, in some cases, on Travis CI. The implementation for version label management from before this PR was indeed complicated, but it worked reliably on many different platforms and, in particular, with older versions of setuptools.

@johnyf can you comment?

slivingston:
I think that the main issue is platforms with older versions of `setuptools` and which have setuptools installed via non-`pip` mechanisms, e.g., as a deb package.

johnyf:
I have not encountered any problems with this approach in other packages, but I use everywhere the latest `setuptools`. An outdated `setuptools` is not very reliable to use, similarly to an outdated `pip`. For example, I think that `pkg_resources` (used in [`version_test`](https://github.com/tulip-control/tulip-control/blob/39d355b04d367eb1d3b26161dc21139cfbc2b6c2/tests/version_test.py#L66)) is missing or elsewhere in `setuptools` from one or two years ago.

A simple solution is to add `setuptools` to `setup_requires`.

Both `setuptools` and `pip` change frequently, and they are improving. The versions in stable branches of Linux distributions like Debian and Ubuntu are (significantly) outdated. A `virtualenv` should be used instead, or python be installed by the user (or both). Reasons along these lines have led me to abandon MacPorts too. All these package-manager (for python) within package manager (`apt-get` or `port`) approaches are nested, and I think that this is why they have more issues than a flat python installation that is disjoint from the system's python (could simply a `virtualenv`).

johnyf:
I am assuming that the Travis failure mentioned is [here](https://travis-ci.org/tulip-control/tulip-control/builds/140310492).

Also, we are upgrading with `pip install -U pip setuptools` [on Travis](https://github.com/tulip-control/tulip-control/blob/5a4b5f1b3cc1808bf73df17cc422778a0b681ccd/.travis.yml#L34).

johnyf:
It seems that `miniconda` contains `setuptools == 23.0.0`, and `pip install -U setuptools` fetches `setuptools == 23.1.0`, which was released yesterday, and then attempts to remove `23.0.0` from a non-existent `easy-install.pth` file at `/home/travis/miniconda/lib/python2.7/site-packages/easy-install.pth`. It may relate to `miniconda`'s packaging. We may want to pin within `.travis.yml` the version of `setuptools` that we require, so that it matches that of the `miniconda` that we fetch.

johnyf:
Also, this particular Travis failure is unrelated to the new version management system, because it happens before `tulip` is involved. It is related to `miniconda`, which was introduced in order to run `scipy` before Travis times out (and w/o increasing the timeout allowance either, to avoid wasting resources). `miniconda` started being used in this same PR, even though orthogonal to version management, in order to get all tests to pass.

slivingston:
Good observation regarding `pip install` vs. `miniconda` on Travis CI. I can try to solve it as part of #151.

slivingston:
[Your comment above](https://github.com/tulip-control/tulip-control/pull/145#issuecomment-228586658) is correct, and my first hypothesis about Tulip version management being related is false.

johnyf:
In case it may be helpful, the only other [occurrence of the same error](https://s3.amazonaws.com/archive.travis-ci.org/jobs/43643285/log.txt) that Google finds.

There appears to be a discussion about the same error, and it solution, for `tensorflow` here: https://github.com/tensorflow/tensorflow/issues/622, and at `anaconda-issues`: https://github.com/ContinuumIO/anaconda-issues/issues/542

slivingston:
I agree with [your comment above](https://github.com/tulip-control/tulip-control/pull/145#issuecomment-228586161) about preferring to use current versions of dependencies, especially critical infrastructure packages like `setuptools`. I am currently using `virtualenv`, which I had used while reviewing this PR. It is superficially easy to switch to use `setuptools` that is installed via `pip`, instead of from a Debian package. However, I had some difficulty both when reviewing this PR and after merging the current master branch in preparation for #151. I have not yet fully diagnosed it, but thus far, it seems to be related to conflicts with other parts of the Python installation that come from Debian packages (and that are out-of-date).

slivingston:
I resolved the issue that [I refer to above](https://github.com/tulip-control/tulip-control/pull/145#issuecomment-228587876). Ultimately the trouble came from an out-of-date `pkg_resources` package that was installed from a Debian package on which other Debian packages depend. Removing such old versions of basic Python packages is challenging because it might require a broken state of the local Debian packages configuration (unless, e.g., a new one-off Debian package is created). For posterity, note that [virtualenv](https://virtualenv.pypa.io) by default does not provide access to system site packages and thus is a solution here. However, if many virtualenvs are used, then there might be repeated installations of basic packages like `numpy`, which are not installed system-wide to avoid breaking the Debian packages configuration.

johnyf:
[`virtualenv-clone`](https://github.com/edwardgeorge/virtualenv-clone) may be useful to skip re-creating identical base virtual environments (though I'm not using it). Multiple installations of the same packages may be tolerable for the isolation that is achieved this way.

johnyf:
I just found [`setuptools_scm`](https://github.com/pypa/setuptools_scm/), which is a `setuptools` extension for the purpose served by the function `setup.git_version`. However, I do not agree with its:

- design
- assumptions (e.g., getting the latest tag) and defaults
- complexity of configuration
- reinventing the wheel (VCS interfaces -- it doesn't appear to depend on widely used packages that specialize on this)

It's design is as nested and complex as `distutils` and `setuptools`, which means notably difficult to guess the corner cases, and fragile. It doesn't seem to be asserting that a version mentioned in code matches that obtained from the VCS.

PR #151, opened 2016-06-26
renewed gridworlds

slivingston:
This pull request re-introduces `tulip.gridworld`, which has been absent since version 1.0a. Examples are under `examples/gridworlds/`. A section of the User's Guide is dedicated to it.

Tests are provided and added to the "base" family. The tests do not involve anything beyond required dependencies and do not excessively increase testing time, so including them in the basic suite is justified.


slivingston:
This pull request introduces new and renewed code; parts that existed previously have not been available on master branch for many months and across several major version changes. Thus, none of the current API is affected (indeed, it is extended by this PR).

slivingston:
@dathath I am aware that you have been working with code as of previous commits in the branch of this pull request. If you have anything that still uses `tulip.gridworld`, can you confirm that your programs still run correctly (given recent changes on this branch)?

johnyf:
Regarding backward compatibility, I think that the API need to change to conform to PEP8. This will affect the methods:

- `isEmpty` -> `is_empty`
- `setOccupied` -> `occupy`
- `setEmpty` -> `free`
- `isReachable` -> `is_reachable`
- `dumpsubworld` -> `dump_subworld`
- `dumpPPartition` -> `dump_ppartition`

(otherwise, it will be a naming situation similar to `logging` within the standard library).

johnyf:
Regarding backward compatibility, I think that the API need to change to conform to PEP8. This will affect the methods:

- `isEmpty` -> `is_empty`
- `setOccupied` -> `occupy`
- `setEmpty` -> `free`
- `isReachable` -> `is_reachable`
- `dumpsubworld` -> `dump_subworld`
- `dumpPPartition` -> `dump_ppartition`

(otherwise, it will be a naming situation similar to `logging` within the standard library).

slivingston:
I agree with your proposed method name changes, except for `setEmpty` and `setOccupied`. How about `mark_empty` and `mark_occupied`?

johnyf:
I agree with those alternatives. I tried to choose some verbs, and noticed that I should have put in parenthesis that these are suggested candidates, but after posting.

slivingston:
Besides asking above, I also sent @dathath an email several days ago. It is not clear whether a response will be received. I think that the review of this pull request can proceed without waiting for one.

dathath:
I did respond three days ago. Was there something more than I responded to?

dathath:
This was my earlier response:

Yup, the code I have works fine as of
commitc408966c87990de03008aa617ef5fe696b31e209

Sumanth

slivingston:
@dathath Thanks. Your original response did not post to this pull request.

johnyf:
@dathath it may work better to respond within a browser in the future.

johnyf:
[PEP8 recommends](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) that instance variables have lowercase names. The variable `tulip.gridworld.GridWorld.W` could be renamed. It does not appear to be used in the examples, and only lightly used in the tests, so this may not be a big change.

In any case, this change can happen after merging into branch `master`, because this branch will not be rebased, in order to preserve history and to avoid potential conflicts during rebasing.

slivingston:
Thanks for reviewing. I agree about changing the name of `W` and will do it later. My original choice was motivated by brevity and following the non-Python style of matrices having identifiers that are single, uppercase letters.

tests/gridworld_test.py
```diff
@@ -279,14 +279,14 @@ def extract_coord_test():
         yield extract_coord_check, label, expected_coord
 
 
-def eq_GridWorld_check(G, H, eq):
```
slivingston: I agree with the name change. For posterity, the previous name was originally motivated to emphasize the relationship with the class `GridWorld` by preserving the letter casing of the class name.

Issue #74, opened 2014-03-19
discretize produces incomplete spatial adjacency matrix


johnyf:
The implementation in `discretize` produces a (symmetric) spatial adjacency matrix, where a 1 signifies that two regions are adjacent. However, the code [updating the adjacency matrix](https://github.com/tulip-control/tulip-control/blob/a7f39d1770719304b92ecaac0881b66bc666b5a2/tulip/abstract/discretization.py#L464) seems to not be adding at all adjacencies between regions that result from [separation of the difference `sol[i] \ s0`](https://github.com/tulip-control/tulip-control/blob/a7f39d1770719304b92ecaac0881b66bc666b5a2/tulip/abstract/discretization.py#L435) (start region minus the preset of the target region).

This was discovered by comparing the produced `adj` after discretization with an `adj` created "from scratch" by iterating over all pairs of the abstraction and calling `is_adjacent` (which by the way seems to yield the correct result - as far as visual inspection can tell).

The algorithm used was the extended version on branch `dbg`, which implements correctly the `use_all_horizon` argument passed to `discretize`. To this end it [implements separation of the intersection](https://github.com/tulip-control/tulip-control/blob/3d3383beecce942a6499e3ee0f2d8e953e1ec3b2/tulip/abstract/discretization.py#L661) `sol[i] \cap S0`, in addition to separation of the difference. But the way this patch is implemented does not affect later code in relation to this particular issue of missing adjacency addition.

When time permits I will add a unit test demonstrating the missing adjacencies. This bug is expected to affect any computation which involves separation of the difference set due to disconnectedness.

Separation usually occurs (even by little) due to drift, dynamics, input constraints, or large disturbance, whenever they result in discrete time "jumps" during backward reachability analysis in `abstract.feasible.solve_feasible`. So for systems without drift or dynamics that overwhelm the available control, this will probably cause problems. But otherwise the incomplete adjacency should be readily verifiable as noted above.

Issue #40, opened 2013-12-05
variable length reachability


slivingston:
(Copied from https://sourceforge.net/p/tulip-control/feature-requests/6/ with creation date 2012-02-22.)

In the current abstraction methods when we establish reachability relations, we look for reachability in exactly `N` steps. It is possible to add an option for reachability in up to `N` steps. This information is already computed in closed-loop algorithm, just needs some book-keeping to make it available as an abstraction option (e.g. within a cell in the partition, we need to know which parts reach in how many steps).

COMMENTS from https://sourceforge.net/p/tulip-control/feature-requests/6/:

2012-02-23 by Nok Wongpiromsarn: Actually, in the MATLAB implementation, this is done by default. (It looks for reachability up to `N` steps instead of exactly `N` steps.) There is a parameter `Options.useAllHorizonLength` that you can set when you call the `discretize` function whether you want reachability up to `N` steps (`Options.useAllHorizonLength=true`) or exactly `N` steps (`Options.useAllHorizonLength=false`).

2012-02-23 by anonymous: Thanks for the heads-up Nok! Good to know what is available ;) I just noticed that we have that option in Python as well but, as far as I can tell, the missing part is online input calculation for this type of abstraction.

slivingston:
@johnyf Why did you label this as a "bug"?

necozay:
I suggest shifting the milestone for this issue to a later release. It is probably better to address this when we include more continuous/hybrid closed loop simulation functions (i.e., together with issue #60)

johnyf:
We have had some correspondence with @slivingston about the issues that I had noted in an old email. We decided, as suggested, to address this in the future, and also transfer a comprehensible description here, from that email.

slivingston:
There are also the following action-items, which I will now do in time for inclusion in v1.2.0:

1. Add note in docstring that the case of `use_all_horizon=True` is
still under development and currently unavailable.

2. Raise ValueError if user attempts to call discretize() or any
function in tulip/abstract/feasible.py with `use_all_horizon=True`.

slivingston:
Both of the above action-items are done as of 430567b026f9639d874dfbd93e2c5375bec4883b.

johnyf:
Based on:

- PEP20: flat is better than nested,
- the [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy): build sharp tools that do one thing well,
[...]

PR #153, opened 2016-07-21
Use all horizon

necozay:
Fixes the horizon length in ``closed loop`` as in the earlier versions of `TuLiP`; and makes the former implementation available under the ``use all horizon`` option.


johnyf:
The docstring of function `feasible.solve_closed_loop` needs to be updated.

johnyf:
Also, in essence, this is an attractor computation (whether exactly depends on the details).

johnyf:
Except for [a union in case of overlap](https://github.com/ajwagen/tulip-control/blob/6f33ca7f7d17332eceb94b199fd78231d469c5c3/tulip/abstract/feasible.py#L183) that does not have any effect in the case of `use_all_horizon == True` (the topic of this PR), this the same algorithm that I (@johnyf) had implemented starting in 55a436712c6632cbdad13aed9271ea0e1e3366db, and which was made unavailable in 430567b026f9639d874dfbd93e2c5375bec4883b.

The reasons that motivated 430567b026f9639d874dfbd93e2c5375bec4883b are not addressed.
Some of those reasons may be missing from the below list, and be commented in the future.

As far as I can understand, the following issues exist with this change:

- timing semantics wrt environment change from fixed horizon (based on a discussion with @necozay).
  The concretized strategy would thus allow the system to choose itself a (potentially different every time the same transition is taken between two given polytopes) horizon during execution. Concretization relies on the assumption that the discrete-valued environment variables remain unchanged between sampling points, i.e., at the horizon boundaries. If the horizon varies during execution, then the system can select the sampling points in time that discrete environment variables will change, which makes the assumption quite different.

- sampling frequency: if the horizon is smaller than the smallest time interval that env vars are known to not change, then at first it appears that the system would pick a higher sampling frequency, so env wouldn't change anyway in between. However, the problem is that the sampling *time instances* shift, and thus can skip environment changes at a lower frequency.

  For example, if env changes at multiples of 4 sec: `*0* 1 2 3 *4* 5 6 7 *8*`, and sys is allowed to pick a horizon 1 <= H <= 4, then it can pick 2 and then 4, to observe the environment at times 1, 2, and 6: `^0^ 1 ^2^ 3 4 5 ^6^ 7 8`. Combining observation and actual env changes, we get `^*0*^ 1 ^2^ 3 *4* 5 ^6^ 7 *8*`. So, sys was traveling from one polytope to another, under the assumption that env discrete vars remain unchanged, during the times 4--6, even though env *did* change. In other words, the synthesized strategy cannot any more be concretized and ensure the given specification at the finer time scale.

- the above cannot be avoided by just stuttering at the destination, after reaching it. Depending on the control input and drift `K` in `x' = A x + B u + E d + K`, even w/o noise, it can be impossible to stutter.

- connectedness: this algorithm, which is (almost) that by @johnyf (me), doesn't work correctly. In case of sufficient drift / strong dynamics that the system "jumps" between polytopes that are disjoint, the package `polytope` does not work correctly. Operations between polytopes return incorrect results. (At least that is my recollection. Relevant issue must have been #74).

  So, firstly `polytope` needs to be corrected and unit tested to ensure that it supports regions that comprise of disconnected (non-)convex polytopes. Please note that this issue doesn't arise with `use_all_horizon == False`, because there we have a chain of polytopes, and each is ensured to be connected, by induction, starting from the connectedness of the final target polytope.

  In summary: polytope connectedness is a loop invariant of the algorithm for `use_all_horizon == False`. It is not an invariant of the algorithm for `use_all_horizon == True`. Given that `polytope` doesn't support (at least as far I have experienced in 2014) disconnected regions, it cannot be used in a loop that does not support this invariant.

- variables `reached` and `old_reached` are ineffective, an error introduced by myself @johnyf in 6a8e8002ab681abae687579ecad245976affb56f, where `reached` is introduced for early termination, but not updated.


Relevant versions (listed here for easy reference):

- [Old version before @johnyf changed it](https://github.com/tulip-control/tulip-control/blob/3b5e538d39fc264af279ff008e12502e1ce992e9/tulip/abstract/feasible.py#L115)

```
p2 = target
for ...:
    s0 = solver_open_loop(p2, N=1)
    p2 = s0 \/ p2
    if dim(p2) < ambient:
        break
return p2
```

- [version after @johnyf modified it](https://github.com/tulip-control/tulip-control/blob/55a436712c6632cbdad13aed9271ea0e1e3366db/tulip/abstract/feasible.py#L115)

```
p2 = target
s0 = empty
for ...:
    p2 = solver_open_loop(p2, N=1)
    s0 = s0 \/ p2
    if dim(p2) < ambient
        break
return s0
```

- [version proposed here](https://github.com/ajwagen/tulip-control/blob/6f33ca7f7d17332eceb94b199fd78231d469c5c3/tulip/abstract/feasible.py#L122)

same as previous version, except for an "overlaps initial set?" that has been commented, and does not change the algorithm for the case `use_all_horizon == False`.

Connectedness and drift were some main reasons for deactivating it in 430567b026f9639d874dfbd93e2c5375bec4883b.

Also, after further analysis, the commit message of 55a436712c6632cbdad13aed9271ea0e1e3366db may be incorrect, in that the first and second versions are different, but may be equivalent. This needs searching for what bug led to writing the message in 55a436712c6632cbdad13aed9271ea0e1e3366db.

johnyf:
Relevant issue is #40, where it was decided to deactivate the option `use_all_horizon`. The above comment (partially) answers [this question](https://github.com/tulip-control/tulip-control/issues/40#issuecomment-149130432).

necozay:
This change just cleans up the flags and makes them consistent with what is documented in various publications. With this change:
- flag `closed-loop = True` works with a fixed horizon (as has always been in TuLiP and documented in various publications until the "recent" change by @johnyf  - this change is called @johnyf version from now on). 
- the @johnyf version is now under `use_all_horizon = True`. This has issues in terms of environment timing assumptions as discussed above. 
- as for drift and connectedness, I think appropriate flag/input to play around to fix them is `trans_length`, which is related to whether one only wants to look at the neighboring polytopes to define transitions (where neighboring is defined in terms of adjacency) and up to what length neighbors. looking at the neighbors is motivated by the fact that if there is an underlying continuous-time system, it should not jump. but for discrete-time systems being a neighbor is not really relevant. One option could be to define `trans_length = Inf` that does not use polytope adjacency at all and just looks at all pairs of cells for the case with drift.

johnyf:
Updated the comment [above](https://github.com/tulip-control/tulip-control/pull/153#issuecomment-235748833): for `use_all_horizon`, the algorithms of 55a436712c6632cbdad13aed9271ea0e1e3366db and 92f717f4b381714f8edcb659b27f3b9791a57904 are different.

johnyf:
After email correspondence with @necozay and @slivingston the issue has been investigated and resolved on branch `fixed_or_variable_horizon`, and the conclusion summarized below.


1. Original version, after some refactoring by Ioannis, but w/o any changes to the algorithm.

  Original version, after refactoring, but before the algorithm was modified by @johnyf. For `use_all_horizon`, this is an attractor computation. Depending on the polytopes, it can be conservative in some cases (an under-approximation). For `not use_all_horizon`, this is simply a backward iteration w/o any unions.

  https://github.com/tulip-control/tulip-control/blob/3b5e538d39fc264af279ff008e12502e1ce992e9/tulip/abstract/feasible.py#L115

  ```python
  def solve_closed_loop(
          P1, P2, ssys, N,
          use_all_horizon=False, trans_set=None):
      p1 = P1.copy() # Initial set
      p2 = P2.copy() # Terminal set
      if trans_set is not None:
          Pinit = trans_set
      else:
          Pinit = p1
      # backwards in time
      for i in xrange(N, 0, -1):
          # first step from P1
          if i == 1:
              Pinit = p1
          s0 = solve_open_loop(Pinit, p2, ssys, 1, trans_set)
          p2 = union_or_chain(s0, p2, use_all_horizon)
          # empty target polytope ?
          if not pc.is_fulldim(p2):
              return pc.Polytope()
      return p2


  def union_or_chain(s0, p2, use_all_horizon):
      if use_all_horizon:
          p2 = s0.union(p2, check_convex=True)
      else:
          p2 = s0
      return p2
  ```

2. After @johnyf modified the case `use_all_horizon`, and either intentionally or not (does not
remember) modified the case `not use_all_horizon` too.

  - This change replaced the attractor computation for `use_all_horizon` with a more conservative one, in an attempt to circumvent computation of CPre with a disconnected polytope as target. This change still had issues with union in `polytope` producing a disconnected polytope. See also #74. Variable `p2` here takes the same values as it does in No.1 when `not use_all_horizon` (there).

  - Another reason for this change, which likely triggered it was that No.1 includes states that are already at the target `p2` (0 steps away). But, it may be impossible to go from these states to `p2` in any number of steps from `1..\infty`. Though it is possible in 0 steps (0 time, Zeno behavior).

    This was a bug in the original code. It affected computations where the dynamics includes either a strong constant term, or has a lower bound on the control input `u` (and so the system cannot stay at one place w/o moving).

  - By applying this change, it appears that @johnyf used the same computation also for the case `not use_all_horizon`, possibly because it escaped attention, forgetting to return `p2`, instead of `s0` in that case. Alternative reason could be
because he did not consider the change in sampling assumption of env by sys.
  
 https://github.com/tulip-control/tulip-control/blob/55a436712c6632cbdad13aed9271ea0e1e3366db/tulip/abstract/feasible.py#L115
  
  ```python
  def solve_closed_loop(
          P1, P2, ssys, N,
          use_all_horizon=False, trans_set=None):
      p1 = P1.copy() # Initial set
      p2 = P2.copy() # Terminal set
      if trans_set is not None:
          Pinit = trans_set
      else:
          Pinit = p1
      # backwards in time
      s0 = pc.Region()
      for i in xrange(N, 0, -1):
          # first step from P1
          if i == 1:
              Pinit = p1
          p2 = solve_open_loop(Pinit, p2, ssys, 1, trans_set)
          s0 = s0.union(p2, check_convex=True)
          # empty target polytope ?
          if not pc.is_fulldim(p2):
              break
      #if use_all_horizon:
      #    return s0
      #else:
      #    return p2
      if not pc.is_fulldim(s0):
          return pc.Polytope()
      pc.reduce(s0)
      return s0
  ```

3. Proposed in this PR, to revert the case `not use_all_horizon`.

  What this does is:
  
  - restores the case `not use_all_horizon` to No.1, and
  - leaves the case `use_all_horizon` as @johnyf modified it in No.2.

  https://github.com/ajwagen/tulip-control/blob/6f33ca7f7d17332eceb94b199fd78231d469c5c3/tulip/abstract/feasible.py#L122

  ```python
  def solve_closed_loop(
          P1, P2, ssys, N,
          use_all_horizon=False, trans_set=None):
      # if use_all_horizon:
      #    raise ValueError('solve_closed_loop() with use_all_horizon=True '
      #                     'is still under development\nand currently '
      #                     'unavailable.')
      p1 = P1.copy() # Initial set
      p2 = P2.copy() # Terminal set
      if trans_set is not None:
          Pinit = trans_set
      else:
          Pinit = p1
      # backwards in time
      s0 = pc.Region()
      reached = False
      for i in xrange(N, 0, -1):
          # first step from P1
          if i == 1:
              Pinit = p1
          p2 = solve_open_loop(Pinit, p2, ssys, 1, trans_set)
          if use_all_horizon:
              s0 = s0.union(p2, check_convex=True)
              s0 = pc.reduce(s0)
          else:
              p2 = pc.reduce(p2)
             s0 = p2
          # empty target polytope ?
          if not pc.is_fulldim(p2):
              break
          old_reached = reached
          # overlaps initial set ?
          # if p1.intersect(p2):
          #    s0 = s0.union(p2, check_convex=True)
          #    s0 = pc.reduce(s0)
          # we went past it -> don't continue
          if old_reached is True and reached is False:
              logger.info('stopped intersecting si')
              #break
          if reached is True:
              break
      if not pc.is_fulldim(s0):
          return pc.Polytope()
      s0 = pc.reduce(s0)
      return s0
```

4. As prepared on branch [`fixed_or_variable_horizon`](https://github.com/tulip-control/
tulip-control/commits/fixed_or_variable_horizon) by @johnyf, in order to close this PR. Almost reverts the case `use_all_horizon` to its original version in No.1, thus restoring the attractor computation, while excluding states that are 0 steps away from the target `p2` (by returning `r`, instead of `p2` in the function `underapproximate_attractor`).

  https://github.com/tulip-control/tulip-control/blob/458ea970ff18164e5ade3fb2b25f4f6ad9ec6291/tulip/abstract/feasible.py#L121

  ```python
  def solve_closed_loop_fixed_horizon(
          P1, P2, ssys, N, trans_set=None):
      p1 = P1.copy()  # initial set
      p2 = P2.copy()  # terminal set
      if trans_set is None:
          pinit = p1
      else:
          pinit = trans_set
      # backwards in time
      for i in xrange(N, 0, -1):
          # first step from P1
          if i == 1:
              pinit = p1
          p2 = solve_open_loop(pinit, p2, ssys, 1, trans_set)
          p2 = pc.reduce(p2)
          if not pc.is_fulldim(p2):
              return pc.Polytope()
      return p2


  def underapproximate_attractor(
          P1, P2, ssys, N, trans_set=None):
      """Under-approximate N-step attractor of polytope P2.
      See docstring of function `solve_closed_loop_fixed_horizon`
      for details.
      """
      _print_horizon_warning()
      p1 = P1.copy()  # initial set
      p2 = P2.copy()  # terminal set
      if trans_set is None:
          pinit = p1
      else:
          pinit = trans_set
      # backwards in time
      for i in xrange(N, 0, -1):
          # first step from P1
          if i == 1:
              pinit = p1
          r = solve_open_loop(pinit, p2, ssys, 1, trans_set)
          p2 = p2.union(r, check_convex=True)
          p2 = pc.reduce(p2)
          # empty target polytope ?
          if not pc.is_fulldim(p2):
              return pc.Polytope()
      return r
```

Other comments:

- It is commented [above](https://github.com/tulip-control/tulip-control/pull/153#issuecomment-235764424) that a continuous-time system does not jump (due to continuity of the flow), implicitly suggesting that its discretization should not jump either. The latter is subject to picking a sufficiently small sampling period, i.e., having sufficiently good sensors and actuators. If this is not the case, then jumps (and so loss of connectedness) can arise, which could indicate poor design choices. We may want to detect and warn the user about disconnectedness, so that they reconsider their design choices.

- In the initial response [above](https://github.com/tulip-control/tulip-control/pull/153#issuecomment-235748833), I focused only on the case `use_all_horizon`, because of the issue's title, and so I did not notice that the case `not use_all_horizon` had been affected by my original modification, and was reverted by this PR.

- Thanks also to @rmattila for asking a good question in 2014, which prompted me to record the situation in an email then, and that proved helpful now.

johnyf:
A test for checking the 0-step transitions mentioned [above](https://github.com/tulip-control/tulip-control/pull/153#issuecomment-240142270) had been added (by me) in 140d498aba82e4dfbbff7d2c8818af9a0f7931a2 (2014).

I now added in 23c937ae81785a573a75f93cb6cb5fc2394c743f a test to distinguish between `use_all_horizon` and `not use_all_horizon`.

These two tests very clearly define the issues discussed, and record those that are violated by each version mentioned above. In particular (functions in `tests/abstract_test.py`):


|         |     self-loop   |   `1..N` vs exactly `N` steps   |
|:-------------:|:-------------:|:-----:|
|  Test function | `test_transient_regions` | `test_is_feasible` |
| 23c937ae81785a573a75f93cb6cb5fc2394c743f  | Pass | Fail |
| 989d18f2274ec0cd587010b07c5e86943f084980 | Pass | Pass |
| 0c8f8b8651b9d6e66be509ca3981893f2696eab7 | Fail | Pass |
| d358c2de41a01efa1ce177cc239d9c4027390785 | Pass | Pass |

Merged by rebasing branch `fixed_or_variable_horizon`, with last commit 06a5fc5016fc9a295802d307df4d97fc06aedba8.

PR #155, opened 2016-07-29
Changelog

johnyf:
Added back from pre-v1 and updated with messages found on the GitHub releases page and the `tulip-control-discuss` mailing list.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,62 @@
 # tulip changelog
 
 
+## 1.1a
+2014-12-09
+
+There has not been a release in a while, yet the project has been active.
+We collect here several of the most significant changes,
+though be warned there is more.
+A good place to begin is the installation instructions at
+http://tulip-control.sourceforge.net/doc/install.html
+or fetch the release from
+http://sourceforge.net/projects/tulip-control/files/
+
+ - `tulip.polytope` is now a separate Python package,
+   which is named simply `polytope`. The primary repository for it is https://github.com/tulip-control/polytope
+   and releases are available from PyPI at
+   https://pypi.python.org/pypi/polytope
+
+- support for switched systems with external inputs is introduced
+
+- multiprocessing can be used for discretization of switched systems
+
+- redesign of abstraction (`AbstractPwa`, `AbstractSwitched`) and
+  partition classes (`PropPreservingPartition` and
+  parents that now live in `polytope`)
+
+- `PropPreservingPartition` can now check several of its properties
+
+- support for open finite transition systems in synth
+
+- support for arbitrary finite domains when using the `gr1c` solver
+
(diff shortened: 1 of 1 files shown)


slivingston:
To be sure that I understand your proposed format, for each entry, it is
```
## VERSION
DATE (as YYYY-MM-DD)

...arbitrary text, copied mostly or fully from the announcement email for VERSION...
```
The first (i.e. oldest) several entries do not follow this format.

slivingston:
There is a practice used in other projects to have someone sign the entry for a changelog, using a line of the form
```
Frodo Baggins <fbaggins@shire.net> Fri, 29 Jul 1110 10:12:48 -0400
```
Arguments for:
* there is one person to contact about the release; of course, other humans can be consulted, but that particular person is a best guess for first contact.
* the signature creates some context about the time of release: who updated the changelog and what was the timestamp. These details can be hard to discover later.

Arguments against:
* there might be some confusion about that person's role in making the version; e.g., others might have contributed substantially, but there names might be missed compared to the person who updates the changelog.
* exposing personal email addresses can be excessive to some contributors; after all, we have instead project-level mailing lists etc.

slivingston:
This pull request is good and ready to merge, but I will wait for 7 hours in case anyone wants to comment on my above comments.

johnyf:
Regarding person identities, I generally favor the simplest format possible, because it is mainly consulted by users who quickly skim it, or run into an unexpected new problem. For more information, they will most likely look at the commit history and come to the issue tracker.

About including a single person name: multiple persons can commit different parts of a single changelog message, so that would lead to a list of names. The version tag author is assumed to be the person who uploaded the release, etc.

I am opposed to listing emails in a changelong.

[In `dd`](https://github.com/johnyf/dd/blob/a53bc3214fa392498184a73be8f0e73f7f461615/CHANGES.md), I am using a slight refinement of this format, with a list of general changes, followed by "API:" and a hierarchical list of changes, structured per subpackage, module, function, class, method, etc., so that I won't have to repeat dot-separated names like `tulip.spec.form.GRSpec.__init__` changed, and `tulip.spec.form.GRSpec.__or__` changed, etc.

Yes, the proposed format is

```
## version identifier
YYYY-MM-DD
``

I would omit the date (I do so in `dd`), but I included it only for practical reasons, i.e., as a reference that doesn't need one to go to the GitHub releases page or issue a git command to list tags that match `v*` with their dates.

For the oldest entries, I kept them mostly as they are, to let us decide how we want them formatted. They appear to record some history that may be in the commit history. In general, we can later go back and attempt to fill the missing entries, and also reformat the first entry.

slivingston:
Thanks for your arguments. I support them all for TuLiP, but I want to emphasize that including the date is important.

Issue #159, opened 2016-07-29
distinguish between formulae and synthesis problems


johnyf:
Notionally or in code too. The issue arises when a graph-based state machine description (transition system) is translated to an assume-guarantee GR(1) formula. This is a just a collection of predicates `EnvInit, EnvNext, EnvRecur_j` and `SysInit, SysNext, SysRecur_i`, to be used for creating a temporal formula. It doesn't say any of:

1. what capabilities the strategy has (`GRSpec.moore`)
2. how strict the implication is (`GRSpec.plus_one`)
3. how initial conditions are quantified (`GRSpec.qinit`)

On the other hand, `GRSpec.__or__` raises an exception if its operands disagree on any of the above attributes (otherwise we'll mix mismatched definitions of synthesis(.

Although explicit, it is somewhat impractical to require that an extra arg to `synth.synthesize` define the above for the spec resulting from the graph. It would also *necessarily* duplicate the existing choices in `spec` (otherwise an `Exception` would be raised). So, there is little point in requiring more than one definition of these attributes. There's only one per synthesis problem.

In addition, it would be notionally wrong to add such attributes to a graph-based description of a state machine. That's just a pair of predicates (`Init, Next`), to be used as deemed appropriate (for a state machine in verification, to assemble A/G properties, etc.).

So, my choice is to use the attributes of the `spec` given to `synth.synthesize` to define (more precisely, overwrite the defaults from instantiation of `GRSpec`) these attributes for the `GRSpec` returned by functions `synth.sys_to_spec` and `synth.env_to_spec`.
Please let me know if there are any objections to the above thinking.

Sidenote: in the future, when `omega.automata` and `omega.symbolic.logicizer` will start serving, and eventually replacing, most of `tulip.transys` and the top (long) part of `tulip.synth`, the above two concerns will be more clearly separated, because synthesis interfaces and flattening of graphs to logic will live in different modules.

slivingston:
I understand that this is closed via PR #163. However, this issue is discussion-oriented. @johnyf did you want to keep this open?

johnyf:
I think that this can be closed, because it was implemented in 3208fc1c69787ec45d1e17afaf89a3f0d726ba06 and 596599edb07fe4c7b5f3fbd70e046aa69be4b900.

Issue #156, opened 2016-07-29
simplify API for defining quantification of initial values


johnyf:
I am working on how the user will define how they want the values of variables to be quantified in the initial state.

(Strictly speaking, the quantification is over constants, plus an equality constraint to flexible variables, as in `\E u: x = u`, where `x` has been declared as a (flexible) variable, and `u` is just a local dummy rigid variable declared in the scope of `\E`.)

In `omega`, I've used the [option `qinit`](https://github.com/johnyf/omega/blob/fc55c32c40e0d88e38e905f31919de259b6a7fb0/examples/symbolic.py#L49) passed to enumeration. However, synthesis supports only `\A \A`, so it doesn't have an option (it supported `\E \A`, then I replaced it with `\A \A` for simplicity, and because I believe that `\E \A` / `\A \E` should probably be used less in the future),
Nonetheless, I may add this option again in 0.0.9, until it the one way this should be done [becomes obvious](https://www.python.org/dev/peps/pep-0020/).

To concentrate everything at one place, and make `synth` oblivious to all these options (when arguments start forming long lines, it is time to erase them and add attributes to a data structure), I added the attribute `GRSpec.qinit` (quantification of initial values of variables).
There are 4 values for this option:

- `\A \A`: forall env, sys
- `\A \E`: forall env exist sys
- `\E \A`: exist sys forall env
- `\E \E`: exist env, sys

However, I think that all this can be significantly simplified. (The options are explicit, so that's good, but there's a simple implicit way to define this, and simple is good too.) I arrived at this conclusion from a very different concern, but the result is close to how the options of `gr1c -n` are defined. There is a mathematical reason for their definition.

Based on the definition of synthesis, there's no point in requiring any initial condition at all as guarantee, unless either of 2 things happen (correct me if wrong):

1. the predicate same or stronger to the required init appears as initial condition on the left, i.e., as an initial condition of the realization, or
2. some variables (until now these have always been sys vars, so lets talk about that only) are existentially quantified.

No.1 is mathematically equivalent to conjoining into `Init == env_init /\ sys_init`. So, just leave `sys_init` empty if that's the case, and put everything in `env_init`.
Then, either:

- `env_init` is non-empty and `sys_init` empty, signifying that all vars are universally quantified, or
- `env_init` is non-empty and `sys_init` non-empty, signifying that env vars universally quantified and sys vars existentially quantified (order `\E \A` vs `\A \E` selected based on the attribute `GRSpec.moore`), or
[...]

Issue #131, opened 2015-10-19
Expand Travis CI configuration to perform all tests


slivingston:
https://travis-ci.org/tulip-control/tulip-control

Currently (as of 54d122fbd76a0188f138521628c53132b2aea8ce), only the "base" family of tests is performed. We should expand the CI configuration to perform all tests, i.e.,
```sh
./run_tests.py full
```

slivingston:
While addressing this issue, also improve the manner of getting scipy. Currently it sometimes requires so much time that the build fails due to timeout.

johnyf:
`conda` has been used in 7e9c4b42f29100b85dda7433454116058e25db0b to install `scipy` in less time.

johnyf:
As of b644de60e0a24e48f65fba925dccbb8443c46a6a, all tests for supported features pass, with pure Python support for GR(1) (and Rabin(1)) synthesis. The omitted tests are for:

- `tulip.interfaces.jtlv` (to be completely removed in the future)
- `tulip.interfaces.lily` (experimental, not a production feature)
- `tulip.interfaces.slugs` (outdated, unmaintained, depending on a module available via a fork, more functionality available via other interfaces, and in pure Python)

The `full` family of tests includes these, so a new test family `ci` has been created in `run_tests.py`.

slivingston:
I did not know that there was a distinction with some features being "production". I think that if it is tested in the `full` family, then it should be tested as part of CI jobs. This issue is not addressed until that is the case, so there is no need for a special `ci` family.

slivingston:
The relevant issue about completely removing `tulip.interfaces.jtlv` in the future (as described [above](https://github.com/tulip-control/tulip-control/issues/131#issuecomment-238011328)) is #158.

johnyf:
fa52b12d3bfc4bd5e54b7249f03526377fc6a6b4 tests also `jtlv` and `lily` on Travis CI.
The `slugs` test remains (discussed below).

In https://github.com/tulip-control/tulip-control/issues/154#issuecomment-236415425 the "all tests" was intended to refer to the optional but maintained features. Therefore I misunderstood the OP of this issue.

I will not be able to maintain `tulip.interfaces.slugs`. I do not know whether it runs with the current version of `slugs` and the Python interface available with `slugs`. If the interface can be maintained, then the corresponding test could also be updated, if it needs changes. If being unmaintained would imply that this interface be deprecated / removed, then I do not object to its removal.

[...]

PR #163, opened 2016-08-17
`omega` as default synthesizer, `GRSpec` attributes for synthesis

### LINKED ISSUE #150: enumerate strategy using a graph search
Currently, `interfaces.omega` uses a naive enumeration of the entire relation (by [calling `omega.symbolic.enumeration.relation_to_graph`](https://github.com/tulip-control/tulip-control/blob/f3a0f44e44b52aba1a2448492a556b3f98f58744/tulip/interfaces/omega.py#L60)). The reason is that `relation_to_graph` does a simple naive enumeration of the entire relation, because its intended use is for debugging. It was used in `tulip.interfaces.omega` because it was available at the time of initial writing.

Example where this issue shows up: plotting [`examples/robot_planning/controlled_switching.py`](https://github.com/tulip-control/tulip-control/blob/0590b715f0a40fb0166267c9621768fbacf59cd4/examples/robot_planning/controlled_switching.py) with `dot` never finishes, because the strategy has 769 nodes (as opposed to 23 nodes when using the enumeration of `gr1c`). Simple measurements indicate the difference of naive enumeration vs graph search:

- only 101 of 769 nodes are reachable from `Sinit`
- enumerated relation is full, i.e., at each game state, all possible next moves are available to the controller (some may call this "non-deterministic" strategy -- this term is overused and so unclear)
- a search can *prefer* to choose controller steps that lead to a state that has already been enumerated, as a heuristic for enumerating fewer states.

The search-based enumeration will be implemented within `omega`. The action item here is to change the call within the interface, and the purpose to document the situation. The above originated from email correspondence with @slivingston. This is not a bug, so the `omega` interface can be used.

For reference, within `gr1c`, enumeration [starts here](https://github.com/slivingston/gr1c/blob/388c2515db076be8fc5dc0d2df3e7edf93b68bf1/src/solve.c#L220).

Clearly, the symbolic strategy synthesized can be used directly, without any enumeration. This is necessary for larger examples anyway (that cannot be enumerated in practice), but is not a reason to not improve enumeration. Enumeration is useful for debugging and development (not so much for production).

johnyf:
This PR closes #150, #156, #159. The primary changes are:

- update strategy enumeration in `tulip.interfaces.omega` and use `omega >= 0.0.9 < 0.1.0`
- offer `is_realizable` for `interfaces.omega` (and in `synth` too)
- require `omega` as a dependency, installed from PyPI
- replace `gr1c` with `omega` in most tests, and in examples
- clean, address minor bugs, and uniformize style in examples
- update docs to reflect default synthesizer
- modernize docs about installation
- update problem formulation in tutorial, paying attention to first-order specs, and more correct assume-guarantee notation (see below)
- install `dd.cudd` on Travis CI to test all options available in `interfaces.omega`
- API: add attributes `GRSpec.moore, plus_one, qinit` in order to define between different synthesis problems, different forms of assume-guarantee specifications, and choices for quantification of initial conditions (PEP 20: explicit is better than implicit)
- use the aforementioned attributes in synthesis interfaces, tests, and examples.

There are a number of changes in the documentation. A more detailed discussion could be carried about these, but in summary, some main points are:

- we have first-order specifications, so talking about propositions only does not represent `tulip`. (Talking about propositions doesn't represent an untyped logic either, though for the moment we are still in transition to becoming untyped).
- Writing `=>` in assume-guarantee specifications should be avoided. Use `\overset{sr}{\rightarrow}` instead.
- Using primed variables, instead of "next" and "X" is an approach closer to the tools, papers, and simple ways of writing difference equations (that don't mention time).

It had been agreed that I would merge these changes directly into `master`. However, the changes are in some cases extensive, so a review would guard against having forgotten something, or discussion regarding some design choices.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -1,59 +1,52 @@
 Installation
 ------------
 
-The latest release of TuLiP can be downloaded from `SourceForge
-<http://sourceforge.net/projects/tulip-control/files/>`_.
-
-TuLiP is designed to work with Python version 2.7, though it should also support
-Python version 3.2+.  The following additional Python packages are required to
-use the core functionality of TuLiP, and are installed automatically
-from PyPI when installing with ``pip install tulip``:
-
-* `NumPy <http://numpy.org/>`_
-* `SciPy <http://www.scipy.org/>`_
-* `NetworkX <http://networkx.lanl.gov/>`_
-* `pydot <https://github.com/erocarrera/pydot>`_
-* `PLY <http://www.dabeaz.com/ply/>`_
-* `polytope <https://pypi.python.org/pypi/polytope>`_ -- computations on and
-  plotting of convex polytopes
-
-Newcomers to scientific computing with Python should read
-:ref:`newbie-scipy-sec-label`.
-
-The default synthesis tool for GR(1) specifications is `gr1c
-<http://scottman.net/2012/gr1c>`_. Please install at least version 0.9.0. If you
-do not already have it, the introduction of `the manual of gr1c
-<https://tulip-control.github.io/gr1c/>`_ is a good place to begin.
-
-The following are optional dependencies, listed with a summary of dependent
-features:
-
-* `Matplotlib <http://matplotlib.org/>`_ -- many visualization features
-
-* `Graphviz <http://www.graphviz.org/>`_ -- generation of images (e.g., PNG
-  files) from dot code
-
-* `CVXOPT <http://cvxopt.org/>`_ -- construction and manipulation of discrete
-  abstractions
-
-For computing discrete abstractions from hybrid system descriptions, it is
-highly recommended---but not required---that you install `GLPK
-<http://www.gnu.org/s/glpk/>`_ (a fast linear programming solver). Note that you
-need to install GLPK *before* installing CVXOPT and follow the instructions in
-CVXOPT installation to ensure it recognizes GLPK as a solver. If you are a
-`MacPorts <http://www.macports.org/>`_ user, please note that MacPorts does not
-do this linking automatically.
-
-Once all of the above preparations are completed, you can install TuLiP::
-
-  $ pip install .
-
-TuLiP may instead be installed `from PyPI <https://pypi.python.org/pypi/tulip>`_::
+TuLiP works with Python version 2.7.
+Install it `from PyPI <https://pypi.python.org/pypi/tulip>`_ with::
 
   $ pip install tulip
 
-The above commands include checking of dependencies and automatic installation
-of missing Python packages. (N.B., not all dependencies are Python packages.)
+or from source::
+
+  $ pip install .
+
+``pip`` installs Python dependencies `automatically
+<https://pip.pypa.io/en/stable/reference/pip_install/#installation-order>`_.
+They are listed in ``install_requires`` within ``setup.py``.
+The only Python packages that you may want to install yourself,
+in order to link them properly are:
+
+* `NumPy <http://numpy.org/>`_
+* `SciPy <http://www.scipy.org/>`_
+
+The following are optional dependencies,
+listed with a summary of dependent features:
+
+* `Matplotlib <http://matplotlib.org/>`_ --
+  many visualization features
+
+* `Graphviz <http://www.graphviz.org/>`_ --
+  to plot graphs, for example discrete state machines
+
+* `CVXOPT <http://cvxopt.org/>`_ --
+  construction and manipulation of discrete abstractions
+
+* `GLPK <http://www.gnu.org/s/glpk/>`_ --
+  fast linear programming solver
+
+For computing discrete abstractions from hybrid system descriptions,
+it is highly recommended that you install both CVXOPT and GLPK.
+Note that you need to install GLPK *before* installing CVXOPT,
+and follow the `CVXOPT installation instructions
+<http://cvxopt.org/install/index.html>_`
+to link CVXOPT to GLPK.
+(If you use
+`MacPorts <http://www.macports.org/>`_,
+please note that MacPorts does not do this linking automatically.)
+
+The latest release of TuLiP can be downloaded also from
+`SourceForge
+<http://sourceforge.net/projects/tulip-control/files/>`_.
 
 
 .. _synt-tools-sec-label:
@@ -61,23 +54,34 @@ of missing Python packages. (N.B., not all dependencies are Python packages.)
 Alternative discrete synthesis tools
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-While gr1c is required, as described in :doc:`install`, TuLiP can use other
-tools for formal synthesis. Those for which an interface is available are listed
-below. Also consult :doc:`specifications` concerning relevant syntax and
-summaries of the specification languages. Generally, direct interfaces are
-defined as modules in the subpackage ``tulip.interfaces``. However, these tools
-can be accessed indirectly by appropriately setting parameters for various
-functions in TuLiP, such as ``tulip.synth.synthesize()``.
+The default synthesis tool for GR(1) and Rabin(1) specifications is
+`omega <https://github.com/johnyf/omega>`_
+(installed by ``pip``).
+
+TuLiP can use other tools for formal synthesis.
+Those for which an interface is available are listed below.
+Also consult :doc:`specifications` concerning relevant syntax and
+summaries of the specification languages.
+Generally, direct interfaces are defined as modules in
+the subpackage ``tulip.interfaces``.
+However, these tools can be accessed indirectly,
+by appropriately setting parameters for various functions in TuLiP,
+such as ``tulip.synth.synthesize()``.
 
 These are *optional dependencies*. TuLiP is useful without having them
 installed, but certain functionality is only available when they are.
 
+
 GR(1)
 `````
 
-* `gr1py <https://github.com/slivingston/gr1py>`_
+* `gr1c <http://scottman.net/2012/gr1c>`_.
+  Please install at least version 0.9.0.
+  The introduction of
+  `the manual of gr1c <https://tulip-control.github.io/gr1c/>`_
+  is a good place to begin.
 
-* `omega <https://github.com/johnyf/omega>`_
+* `gr1py <https://github.com/slivingston/gr1py>`_
 
 * `slugs <https://github.com/LTLMoP/slugs>`_
 
@@ -116,6 +120,9 @@ for developers are provided in the :doc:`dev_guide`.
 New to Python?
 ~~~~~~~~~~~~~~
 
+Newcomers to scientific computing with Python should read
+:ref:`newbie-scipy-sec-label`.
+
 If you don't already use Python for scientific computing, consider using
 `Enthought Python Distribution (EPD) <http://enthought.com>`_ or `Enthought
 Canopy <https://www.enthought.com/products/canopy/>`_. This may make the
@@ -129,6 +136,7 @@ Alternatives to Enthought are listed on the `SciPy installation webpage
 
 EPD seems to work fine on most platforms but if you cannot get it to work, more
 alternative packages for Mac OS X and Microsoft Windows are mentioned below.
+
 
 .. _troubleshoot-sec-label:
 
@@ -212,13 +220,15 @@ NumPy, SciPy, CVXOPT, and Matplotlib for your system, consider trying
 The package of gr1c for Windows still cannot be found. But without this package,
 you can also run most TuLiP functions.
 
+
 Installing other Python dependencies
 ````````````````````````````````````
 
-The command ``pip install ...`` or ``easy_install ...`` will usually suffice. To
+The command ``pip install ...`` will usually suffice. To
 get `PLY <http://www.dabeaz.com/ply/>`_, try::
 
   $ pip install ply
+
 
 .. _venv-pydoc-sec-label:
 
@@ -250,6 +260,7 @@ it, try looking at the transys subpackage by entering::
 .. [#f1] On Unix systems, in particular GNU/Linux and Mac OS X, the
          terminal shell treats ``~`` as a special symbol representing
          the home directory of the current user.
+
 
 remote server installation
 ``````````````````````````
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -2,7 +2,8 @@ Installation
 ------------
 
 TuLiP works with Python version 2.7.
-Install it `from PyPI <https://pypi.python.org/pypi/tulip>`_ with::
+Install it with `pip <https://pip.pypa.io/en/stable/>`_
+from `PyPI <https://pypi.python.org/pypi/tulip>`_ with::
 
   $ pip install tulip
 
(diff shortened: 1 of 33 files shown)


slivingston:
Is the case of no environment variables treated? (This might variously be called an empty environment or closed system synthesis.) E.g., using examples/discrete.py after applying the following patch ends with `AssertionError: set([])`
```diff
diff --git a/examples/discrete.py b/examples/discrete.py
index 5afe7cf..aa5cdea 100644
--- a/examples/discrete.py
+++ b/examples/discrete.py
@@ -81,9 +81,9 @@ sys.states.add('X5', ap={'lot'})
 # the park signal is turned off infinitely often.
 #
 # @environ_section@
-env_vars = {'park'}
+env_vars = {}
 env_init = set()                # empty set
-env_prog = '!park'
+env_prog = set()
 env_safe = set()                # empty set
 # @environ_section_end@
 
@@ -110,7 +110,7 @@ env_safe = set()                # empty set
 sys_vars = {'X0reach'}          # infer the rest from TS
 sys_init = {'X0reach'}
 sys_prog = {'home'}             # []<>home
-sys_safe = {'(X (X0reach) <-> lot) || (X0reach && !park)'}
+sys_safe = {'(X (X0reach) <-> lot) || (X0reach)'}
 sys_prog |= {'X0reach'}
 # @specs_setup_section_end@
```

johnyf:
Yes, the case of no environment is treated in the implemented algorithm, except for the programming error described below (or other unknown bugs).

This is an error in the (recently introduced) module `omega.symbolic.fol`, in the method `Context.exist`. The `AssertionError` is raised because quantification [is called](https://github.com/johnyf/omega/blob/b0fb9f29073528c7a95a81297ca1645322f117cf/omega/games/enumeration.py#L224) over no variables. It should be caught and ignored in the method `exist` (PEP 20: "Unless explicitly silenced."), because in this particular call, it is admissible to have no variables. Even better, the call can be avoided altogether, because no variables allows returning immediately the given BDD (quantification over no variables).

The assertion is raised by the function [`omega.logic.bitvector.bit_table`](https://github.com/johnyf/omega/blob/b0fb9f29073528c7a95a81297ca1645322f117cf/omega/logic/bitvector.py#L202). Raising an `AssertionError` there is correct, and I have encountered it before (possibly while developing enumeration, ). There is little point to call `bit_table` with an empty set of variables, so that usually indicates a programming error, hence the design decision to stick to raising an `AssertionError` (PEP 20: "Errors should never pass silently.").

This issue does not apply to the (symbolic) synthesizer in `omega.games.gr1` (and so wouldn't have been caught earlier), because the solver works directly with Boolean-valued variables, and so calls `dd.bdd.BDD.quantify` (its aliases `forall` and `exist`) which simply [returns early](https://github.com/johnyf/dd/blob/c535dee7a2e0d4c7312c383855c6239e87afcd37/dd/bdd.py#L578) if no variables are given (and similarly in `dd.cudd`).

The error traceback tail is:

```python
/home/user/.virtualenvs/dev/lib/python2.7/site-packages/omega/games/enumeration.pyc in _exist_forall_init(g, fol, aut, umap, keys)
    222     only_env_init = fol.exist(aut.control['sys'], env_init)
    223     u = fol.apply('->', only_env_init, env_init)
--> 224     u = fol.forall(aut.control['env'], u)
    225     assert u != bdd.false
    226     sys_0 = fol.pick(u, full=True, care_vars=aut.control['sys'])

/home/user/.virtualenvs/dev/lib/python2.7/site-packages/omega/symbolic/fol.pyc in forall(self, qvars, u)
    127         """Universally quantify `qvars` in `u`."""
    128         r = self.apply('not', u)
--> 129         r = self.exist(qvars, r)
    130         r = self.apply('not', r)
    131         return r

/home/user/.virtualenvs/dev/lib/python2.7/site-packages/omega/symbolic/fol.pyc in exist(self, qvars, u)
    133     def exist(self, qvars, u):
    134         """Existentially quantify `qvars` in `u`."""
--> 135         qbits = bv.bit_table(qvars, self.vars)
    136         return self.bdd.exist(qbits, u)
    137 

/home/user/.virtualenvs/dev/lib/python2.7/site-packages/omega/logic/bitvector.pyc in bit_table(variables, table)
    200     @rtype: `dict` of `dict`
    201     """
--> 202     assert variables, variables
    203     dout = dict()
    204     keys = {'type', 'dom', 'signed',
```

with `tulip @ 596599edb07fe4c7b5f3fbd70e046aa69be4b900` and `omega == 0.0.9`.

johnyf:
I think that I remembered where I saw essentially the same error, and guarded it: [12 lines below](https://github.com/johnyf/omega/blob/b0fb9f29073528c7a95a81297ca1645322f117cf/omega/symbolic/fol.py#L147) the call to `bit_table` that caused the above, in the function `pick_iter`, which avoids calling `bit_table` with no variables. The error there was corrected in https://github.com/johnyf/omega/commit/d28bd6c108aa9ff089788e2eb3b9a822036b7f2b

Also, a note about repeatedly calling `bit_table`. The code in `omega.symbolic.fol` is relatively new, so not optimized yet (on purpose, to keep it simple until usage indicates ways to improve it).

johnyf:
Corrected in https://github.com/johnyf/omega/commit/003440117d83cdc684e3f78e6d6d4fc2cee1bff0.

tests/synth_test.py
```diff
@@ -623,8 +623,12 @@ def tearDown(self):
...
-                          transys.MealyMachine)
+        g = synth.synthesize("gr1c", self.f_triv)
+        assert isinstance(g, transys.MealyMachine)
 
     def test_unrealizable(self):
         assert synth.synthesize("gr1c", self.trivial_unreachable) is None
+
+
+if __name__ == '__main__':
+    multiple_env_actions_test()
```
slivingston: Why is only this test called here?

slivingston:
Thanks for the explanation. Provided that the next `omega` release includes your change and has version number is less than 0.1.0, we will obtain good results here in TuLiP because [setup.py requires `omega` version less than 0.1.0](https://github.com/tulip-control/tulip-control/blob/596599edb07fe4c7b5f3fbd70e046aa69be4b900/setup.py#L112)

slivingston:
I want to update code under examples/developer/ and examples/gridworlds/ to use `omega` and if the defaults do not succeed, to have `GRSpec.moore = False` and `GRSpec.qinit = '\A \E'`.

@johnyf do you have any arguments to not do so? I already have the changes ready locally and can push to the branch of this pull request when you approve.

tulip/interfaces/gr1c.py
```diff
@@ -430,42 +430,15 @@ def check_realizable(spec, init_option="ALL_ENV_EXIST_SYS_INIT"):
...
-              is possible initially.  If spec.sys_init is nonempty,
-              then the strategy need only choose one initial state
-              satisfying it.
-
-        Consult the U{documentation of gr1c
-        <https://tulip-control.github.io/gr1c/md_spc_format.html#initconditions>}
-        for detailed descriptions.
+    @param spec: specification.
+
+    Consult the U{documentation of gr1c
```
slivingston: Is there motivation to keep this link to the gr1c documentation? It directs to the section about initial conditions, yet the parameter `init_option` is removed in this changset.

tests/synth_test.py
johnyf: In tests I use the `__main__` stanza for running individual tests during debugging, so the test `multiple_env_actions_test` must have been the subject of debugging. Otherwise, I just call `run_tests.py` (in particular I have placed an older version in my `$PATH` to use it in other projects too).

johnyf:
I agree about changing the defaults under `examples/{developer, gridworlds}`. If any efficiency issue arises (unlikely, and also the changes are declared ready [above](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241090160)), then `interfaces.omega` can be invoked with the explicit option to use `dd.cudd`.

This reminds me to note the following: `dd.bdd` (i.e., the pure Python implementation) includes reordering by Rudell's sifting, but reordering is *not* invoked automatically. So, if anyone sees too much memory being consumed, but would still like to use Python, then they may want to try invoking reordering from user code, or from inside the solver.

Eventually I will make reordering dynamic (cf. https://github.com/johnyf/dd/issues/9), and it should be straightforward to do in Python (via a decorator). However, I am waiting until some time that will allow me to study in detail the consequences and ensure that it will work correctly.

tulip/interfaces/gr1c.py
johnyf: There are roughly two reasons:

1. a link from the docstring to the tool docs (doesn't have to be the section about initial conditions) as a reminder and for convenience.

2. a note about initial conditions, because each tool may have different restrictions on what initial conditions are allowed (e.g., `gr1c` raising errors if both initial predicates are non-empty and `ONE_SIDE_INIT` selected). `AssertionError`s are now raised below from within the function `select_options`, so this link could serve as an explanation. Perhaps it would be more appropriate to move the link to the docstring of `select_options`.

slivingston: Because `select_options()` is called within `synthesize()`, and because their relationship is not apparent from their docstrings, I agree about leaving the URL in the docstring of `synthesize()`.

doc/install.rst
```diff
@@ -1,83 +1,88 @@
...
+to link CVXOPT to GLPK.
+(If you use
+`MacPorts <http://www.macports.org/>`_,
+please note that MacPorts does not do this linking automatically.)
 
-The above commands include checking of dependencies and automatic installation
-of missing Python packages. (N.B., not all dependencies are Python packages.)
+The latest release of TuLiP can be downloaded also from
+`SourceForge
+<http://sourceforge.net/projects/tulip-control/files/>`_.
```
slivingston: When I try to visit <http://sourceforge.net/projects/tulip-control/files/>, the HTTP connection is upgraded to HTTPS automatically (thus becoming <https://sourceforge.net/projects/tulip-control/files/>). Is there any motivation to keep the scheme as http and not https here?

johnyf: No, there isn't. Also, I was tempted to delete this sentence entirely, because, now, distribution is primarily via PyPI. Maybe SourceForge offers some redundancy, though it may be disorienting for unfamiliar persons reading the early sections of the installation guide.

slivingston:
Should instructions about installing `dd` with CUDD be added to the Installation page of the User's Guide?

It does not happen automatically, and some tests in tests/omega_interface_test.py will fail without it, in particular: test_synthesis_cudd(), test_is_circular_cudd().

doc/install.rst
slivingston: I think that the redundancy argument is motivating both technically (in terms of service availability) and because having some presence on SourceForge.net might improve visibility. The risk of disorientation remains, so perhaps we should revisit the question after this pull request.

Also, note that the location of files on SourceForge is provided in the landing page of tulip-control.org, which comes from the source file at doc/website/index.html.

slivingston:
I recommend that tests that require gr1c are moved from the basic family to "full". My motivation is that `./run_tests.py` should succeed if all required dependencies are present. If someone is interested in using and testing against gr1c, then she can `./run_tests.py -f gr1cint` or `./run_tests.py full`.

johnyf:
In general, I try to keep documentation close to each tool, to avoid duplication (both for the author and the reader -- so that they won't have to compare multiple sources of information) and ensure that docs remain up to date with the code.

A link could be added to the [relevant section](https://github.com/johnyf/dd/tree/v0.4.0#cython-bindings) in the installation instructions of `dd`.
I improved that section [yesterday](https://github.com/johnyf/dd/tree/dev#dd-fetching-cudd), by explicitly listing the necessary commands, so that it require less participation from the user.
This is only a recommendation, based on my personal preferences as a reader. I am neutral about including instructions about CUDD. We are talking about 4 commands, so this is a minor issue.

A (functional) variant I can think of is to place these 4 commands in a shell file, e.g., `extern/get_dd.sh` and point from the installation docs to that file. In principle, that wouldn't duplicate the instructions in the README of `dd`, but rather implement them (it doesn't make much sense to implement them within the `dd` sources).

johnyf:
I agree about `gr1c` tests, though until #131 is resolved, it would be good to be running the `gr1c` tests explicitly, by invoking them as noted [above](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241224874).

johnyf:
A relevant commend is about those `dd` tests that use CUDD, also mentioned [above](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241224448). These will fail when running the `base` family. So, they should somehow be distinguished. For now the only way I can think of is by placing them in a separate test file, which is not a very elegant solution.

slivingston:
Re https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241225416
if you do not want to move the tests to a separate file, then they can also be distinguished using attributes. E.g., we have the `slow` attribute [applied to transition_directions_test at line 120 of abstract_test.py](https://github.com/tulip-control/tulip-control/blob/06a5fc5016fc9a295802d307df4d97fc06aedba8/tests/abstract_test.py#L120), which causes the test to be skipped if `--fast` is passed to run_tests.py; cf. lines [78-79](https://github.com/tulip-control/tulip-control/blob/06a5fc5016fc9a295802d307df4d97fc06aedba8/run_tests.py#L78-L79) and [129-130](https://github.com/tulip-control/tulip-control/blob/06a5fc5016fc9a295802d307df4d97fc06aedba8/run_tests.py#L129-L130) of run_tests.py.

johnyf:
Yes, I considered that possibility, but would require defining an additional annotation, and would also would allow running only a subset of the `base` tests (strictly speaking). So, moving to a separate test file seems inevitable for now. At least it is explicit. It can named as the original file, with the replacement `_cudd.py` for its suffix.

slivingston:
This is good work. The remaining concerns can be addressed after merging:
* work on #131 can address the comments about [moving the tests for gr1c to full](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241224874) and [separating tests of `dd` that require CUDD](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241225416);
* documentation about installing `dd` with CUDD can be committed directly to master branch; this might only be a pointer to the documentation of `dd` itself, as [commented above](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241224982), or a shell script that is essentially copied from .travis.yml.
* the scripts under examples/developer/ gr1_arbitrary_set.py, switch.py, and sys_and_env_ts.py are broken, but the errors do not appear primarily related to the changes of this pull request. I am working on patches and will commit them tomorrow.

slivingston:
There was a commit on which discussion occurred about using raw string notation for `GRSpec.qinit`. After it was squashed (i.e., combined) with an earlier changeset, I cannot find the comments, but I would expect them to be available with some text preface like "...commented on an outdated diff".

Summary:
* It is not necessary to use raw string notation, e.g., it is OK to use `spc.qinit = '\E \A'` instead of `spc.qinit = r'\E \A'`.
* Not writing the prefix `r` is more concise and less distracting for users who do not know about raw strings, i.e., strings for which escape sequences are not parsed.
* However, using raw strings explicitly captures the intent of the `qinit` string and precludes Python parsing attempts.
* Some packages define special meanings to strings that begin with `'\'`; e.g., `'\A'` is used in [re](https://docs.python.org/3/library/re.html) in the standard library,

johnyf:
The comments still exist, here: https://github.com/tulip-control/tulip-control/commit/eee5bfd1a22b04984e5a037664c0ad884ffb1fb4#commitcomment-18715720

For similar situations: I recovered the link by looking at our public activity feed. Also, I've noticed that github maintains dangling commits, as long as they are mentioned somewhere in the issue tracker.

PR #165, opened 2016-08-29
Update interface with Slugs; reorganize tests

### LINKED ISSUE #131: Expand Travis CI configuration to perform all tests

slivingston:
closes #131.

This pull request builds on the work of @johnyf by reorganizing tests in accordance [with discussion](https://github.com/tulip-control/tulip-control/pull/163#issuecomment-241238329) on PR #163, i.e.,
* tests that depend on gr1c are part of the full, but not the "basic" family; 
* tests of `interfaces.omega` that require `dd` built with CUDD are similarly moved out of the "basic" family.

This pull request also updates `tulip.interfaces.slugs`. A script is provided to obtain it; the script has only been tested on GNU/Linux x86_64.


johnyf:
I already skimmed most of these changes earlier today. I may have called the separate file `omega_interface_cudd_test.py`, though "extra" anticipates potentially other forms of tests in there. For example, tests that use both `gr1c` and `omega`, or tests that use libraries other than CUDD. Also, "extra" aligns well with the meaning of "extras" in the context of `setuptools`. We could rename the file to `omega_interface_extra_test.py` to ensure the two related files share their prefix, thus appearing next to each other in the lexicographic listing of files (which is convenient for the developer and more structured for the reader).

slivingston:
I agree with the motivation about useful lexicographic ordering of files by moving the `extra` prefix to later in the file names. However, similar prefixes implies more text is required to use `./run_tests.py`. E.g., if we rename extra_omega_interface_test.py to omega_interface_extra_test.py, then
```sh
$ /run_tests.py -f omega_interface
Traceback (most recent call last):
  File "./run_tests.py", line 224, in <module>
    main()
  File "./run_tests.py", line 154, in main
    (basename, match))
Exception: ambiguous base name: omega_interface, matches: ['omega_interface_test.py', 'omega_interface_extra_test.py']
```
so it is necessary instead to write `./run_tests.py -f omega_interface_test`. The above example could also be considered an argument in favor of common prefixes because the user is made aware of variants. Thus, I think that both naming schemes are motivated.

slivingston:
I added a commit that fetches gr1c that is already built for Linux x86_64, which is the platform used in the TuLiP configuration on Travis CI. Because it might be of interest later, I recommend keeping the previous commit on this branch in which the gr1c source release is obtained and used to build gr1c.

johnyf:
The binary for `gr1c` is a good idea, we might want to consider whether there can be some analog for `numpy` and `scipy`. Mentioning this, I wonder why `scipy` doesn't install quickly on Travis CI from the PyPI wheel (i.e., without building to link).

About fetching only the contents of a specific commit of `slugs` from its repository, the following does so, without cloning the repository:

```shell
curl -L https://github.com/VerifiableRobotics/slugs/tarball/ad0cf12c14131fc6a20fe29edfe04d9aefd7c6d4 -o slugs.tar.gz
echo '04c8de023bddc6579fe5f1fe495dea641d3e4cf25bd48ff3a6b0a781d968136a6a635e07c9c1fcf0c784df60b72c24aab8029f0c613c71fed9c104f0e9f046ce  slugs.tar.gz' | shasum -a 512 -c -
tar xzf slugs.tar.gz
```

In particular, this will continue to work in the future, independently of whether branch `master` of `slugs` moves forward (whereas `git clone --depth=10 ... git checkout commit_hash` will prompt us to update the ref every time that slugs moves forward by 10 commits, also producing history noise).
This idea originated from [here](https://coderwall.com/p/-zzhba/download-code-snapshots-from-github-without-using-git). My motivation for researching this more is motivated primarily by the general applicability of this method.
Would this approach not suffice?

For those interested in `git`'s design philosophy, a thorough investigation and commentary on why one cannot fetch individual commits can be found [here](http://stackoverflow.com/a/26135822/1959808).

About the path recovery in the function `tulip.interfaces.slugs._call_slugs`, I do understand that it is likely impossible to do otherwise, unless the user has added also the python script's path to the `$PATH` (and it is not so crucial here to justify further changes). But for other use cases, a useful approach that works on both Unix and Windows, and took me some time to find about is to explicitly pass the environment's path to `subprocess` (https://github.com/erocarrera/pydot/commit/bc639e76b214b1795ebd6263680ee55d9d4fca9f).

I did not think about the longer typing. It works though the other way too: there are multiple `extra_*` files (and may increase in number in the future), so one will still need to type 2 words (not 3 though).

johnyf:
Rebased and merged as of 42523acce3b10e3ef698d1211c9ebec154e8c97b.

slivingston:
Thanks for your review. Regarding some of your comments [above](https://github.com/tulip-control/tulip-control/pull/165#issuecomment-244039987):

1. I had considered trying to directly fetch a tarball from GitHub using the commit hash. However, I did not find any official documentation about doing it by GitHub, so I guessed that it might not be reliable. However, checking the SHA 512 hash as you show is a good idea in any case. I am happy with both solutions (`git clone` or `curl`) for now.

2. I had considered manipulating the variable `PATH`, e.g., when using `subprocess`. However, I think that Slugs (as a tool) is somewhat messy, and adding paths to the system `PATH` into the Slugs source tree or installation location, especially for unfortunately named files like "compiler.py", could be a bad security decision.

johnyf:
About fetching a tarball directly from GitHub, the official documentation (which also uses `curl`) appears to be in the GitHub [Developer API](https://developer.github.com/v3/repos/contents/#get-archive-link).

slivingston:
The section about "Get archive link" describes it as "A valid Git reference", which I thought was ambiguous. However, I now notice that `ref` as a parameter in other methods is described as "The name of the commit/branch/tag", e.g., <https://developer.github.com/v3/repos/contents/#get-contents>.

johnyf:
Changed to `curl` in 7647d64607c731f060a223a4e121f5be3ec34678.

slivingston:
Note that the upstream Slugs repository now uses tags. https://github.com/VerifiableRobotics/slugs/issues/16#issuecomment-259097415

Issue #154, opened 2016-07-28
run using scipy instead of cvxopt


johnyf:
Likewise to `polytope`, which can now run w/o `cvxopt`, `tulip.abstract.find_controller` should [not import `cvxopt` only](https://github.com/tulip-control/tulip-control/blob/9aa793a473aeb5f4a656e2fe45112cbea83becda/tulip/abstract/find_controller.py#L51). This is the only place where `cvxopt` is used. Due to `tulip.abstract.__init__`, it causes failure of discretization too, whenever `cvxopt` is absent.

necozay:
Just to make sure we do not replicate work: This is on my agenda for UM TuLiP working meeting tomorrow. @johnyf if you have already started working on it, let me know. Otherwise, if we cannot get to this, I can leave it to you.

johnyf:
Thanks, I haven't yet touched this, please go ahead. The only place that `grep -ilr "cvxopt" ./*` shows modifications are needed is [`tulip.abstract.find_controller`](https://github.com/tulip-control/tulip-control/blob/fc5d04fdcac9dab2104802a77f5a2c718c7e6bf3/tulip/abstract/find_controller.py#L51).

#131 is relevant to thoroughly test this, both with:

- `scipy` present but `cvxopt` absent, and
- `scipy` present and `cvxopt` present *and* linked to `glpk`

For an example, this has already been done in [`polytope`'s `.travis.yml` on branch `dev`](https://github.com/tulip-control/polytope/blob/065f7699ebb9c4b10a520647d138e850b12b80f2/.travis.yml#L39).

johnyf:
For performance reasons, we probably want to run twice (one with `scipy` and once with `cvxopt.glpk`) only a selected subset of tests that is representative of abstraction and continuous feedback control functionality.

necozay:
@sahiny is looking into this.

johnyf:
Thanks for the note. @sahiny, please let us know in case there is anything we can help with / answer (both here and over email work for me).

slivingston:
If I understand this issue correctly, it does not necessarily involve changing the Travis CI script.  I assigned #131 to myself, so I can ensure that tests touch `scipy` and `cvxopt`.

johnyf:
Running all the tests has been done on branch `enumeration` (though they will need `omega >= 0.0.9` to pass on Travis CI. Running with both `scipy` and `cvxopt` has not been done.

Please see:

[...]

Issue #166, opened 2016-09-02
test computation of controller input


johnyf:
This refers to `tulip.abstract.find_controller`. If needed, rewrite. It is labeled as a bug because it has been reported that it does not match what abstraction says.

necozay:
`find_controller` needs some adjustments to make sure it is consistent with the latest changes in the flags of abstraction (`use_all_horizon`, `closed_loop`, etc.).

But other than that, do we have an example demonstrating a bug? I heard people complaining but nobody shared a concrete example and I could not find a buggy example myself.

johnyf:
There is an example that I do not yet have myself, but when I obtain it and investigate, I will try to create a minimal example. The problem reported is that abstraction and synthesis succeed, and a controller is found for horizon length `N = 1`, but increasing to `N > 1` again abstraction and synthesis succeed, but no controller can be found. Unless this is not due to parameter mismatch (which I have not yet confirmed), it must be something else (either missing implementation, or bug in existing).

slivingston:
Should this be targeted after release of version 1.3.0 ?

johnyf:
Yes, I did not intend to suggest otherwise.

Issue #162, opened 2016-08-13
annotate continuous states with N-attractor they belong to


johnyf:
Task item: process a partition resulting from abstraction of a continuous dynamical system to add information about how many steps are needed to reach the target polytope of an (outgoing) trajectory, from each continuous state.

Different states in the same cell (polytope) `p1` can require a different number of steps in order to reach the target polytope `p2`, say from state `s1` we can reach `p2` in 10 steps, but from state `s2` we can reach `p2` in 12 steps, but not in 10 steps.

As proposed in https://github.com/tulip-control/tulip-control/issues/40#issuecomment-37120298, this information can be added by:

1. repeat the continuous attractor computation, this time storing *each* `k`-step attractor, for `k \in 1..N`
2. store this information in some form of data structure (to be decided as part of implementing this issue), for each outgoing transition from a cell, a partition of that cell.

The above is essentially, for each transition between cells, memoization of the `k`-step (continuous) attractor under-approximations for `k \in 1..m`, where `m < N` suffices to cover the starting polytope `p1`.

The ["online input calculation"](https://github.com/tulip-control/tulip-control/issues/40#issue-23817123) can still be performed with the function `abstract.find_controller.get_input` (or a future renamed version of it). What will be missing for simulation or exporting controllers is the logistics of reading the memoized `k`-step attractors, in order to invoke `get_input` with a sufficient horizon length.

In order to separate concerns, the issues of simulation and controller export / deployment / real-world implementation are *not* task items for this issue. Instead, they are to be resolved:

1. simulation in #60
2. export in #69, #59, and possibly a yet-nonexistent issue.

necozay:
I recently noticed after some discussions with @johnyf  that storing the number of steps might be a bit trickier in that when the system is subject to disturbances and attractors involve non-convex sets, certain points cannot be guaranteed to reach the target in exactly k steps but they can reach the target in k +/- 1 steps depending on disturbance realization. If the Pre of non-convex sets are conservatively computed by taking Pre's of each component, then this might not be an issue.

Another related thing is the following. @sahiny recently observed that even when fixed horizon `N` is used in abstraction, the transitions computed with `closed_loop = true` require `N-k` step attractors in online input computation. For a point `x` that can reach a set `P` using the closed loop mpc in `N` steps, solving the open-loop N step reachability and using the first input might not be feasible (if this is feasible, then `x` is in the open-loop reachable set). `x` is only guaranteed to reach `N-1`-step closed loop attractor.

Issue #146, opened 2016-06-04
simulation part of continuous.py


necozay:
It would be nice to demonstrate how the synthesized controllers can be simulated in TuLiP in one of the examples. I have the following piece of code for the ``continuous.py`` but there are some hacky parts. Maybe we can clean it up and include in the next release. Any suggestions on how to make the initialization better?

```
# Simulation

print '\n Simulation starts \n'
T = 100;

# let us pick an environment signal
# from: http://code.activestate.com/recipes/577944-random-binary-list/

from random import *
randParkSignal = [randint(0,1) for b in range(1,T+1)]

# initialization is a bit hacky (need to be consistent with ctrl init
# states)

s0_part = ctrl[ctrl.edges('Sinit')[1][0]][ctrl.edges('Sinit')[1][1]][0]['loc']

init_poly_v = pc.extreme(disc_dynamics.ppp[s0_part][0])
x_init = sum(init_poly_v)/init_poly_v.shape[0]
x = [x_init[0]]
y = [x_init[1]]

N = disc_dynamics.disc_params['N']
s0_part = find_discrete_state([x[0],y[0]],disc_dynamics.ppp)
ctrl = synth.determinize_machine_init(ctrl, {'loc':s0_part})
(s, dum) = ctrl.reaction('Sinit', {'park':randParkSignal[0]})
print dum,'\n'
for i in range(0,T):
    (s, dum) = ctrl.reaction(s, {'park':randParkSignal[i]})
    u = get_input(
            np.array([x[i*N],y[i*N]]),
            sys_dyn,
            disc_dynamics,
            s0_part,
            disc_dynamics.ppp2ts.index(dum['loc']),
            mid_weight=100.0,
            test_result=False)
    for ind in range(N):
        s_now = np.dot(
                    sys_dyn.A, [x[-1],y[-1]]
                    ) + np.dot(sys_dyn.B,u[ind])
        x.append(s_now[0])
        y.append(s_now[1])
    
    s0_part = find_discrete_state([x[-1],y[-1]],disc_dynamics.ppp)
    s0_loc = disc_dynamics.ppp2ts[s0_part]
    print s0_loc, dum['loc']
    print dum,'\n'

show_traj = True
if show_traj:
    import matplotlib.pyplot as plt
    plt.plot(x)
    plt.plot(y)
    plt.show()
```

johnyf:
[...]

PR #168, opened 2016-11-12
Find controller lp

necozay:
Addresses the issue #154 .

### DIFF
--- a/examples/continuous_with_sim.py
+++ b/examples/continuous_with_sim.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python
+"""
+This example is an extension of robot_discrete.py by including continuous
+dynamics with disturbances.
+
+Petter Nilsson (pettni@kth.se)
+August 14, 2011
+
+NO, system and cont. prop definitions based on TuLiP 1.x
+2 Jul, 2013
+NO, TuLiP 1.x discretization
+17 Jul, 2013
+"""
+#
+# Note: This code is commented to allow components to be extracted into
+# the tutorial that is part of the users manual.  Comments containing
+# strings of the form @label@ are used for this purpose.
+
+import logging
+logging.basicConfig(level=logging.INFO)
+
+# @import_section@
+import numpy as np
+import polytope as pc
+from tulip import spec, synth, hybrid
+from polytope import box2poly
+from tulip.abstract import prop2part, discretize
+from tulip.abstract.plot import plot_partition
+from tulip.abstract.find_controller import *
+#import find_controller_LP
+# @import_section_end@
+show = False
+
+# @dynamics_section@
+# Problem parameters
+input_bound = 10.0
+uncertainty = 0.01
+
+# Continuous state space
+cont_state_space = box2poly([[0., 3.], [0., 2.]])
+
+# Continuous dynamics
+A = np.array([[1.0, 0.], [ 0., 1.0]])
+B = np.array([[0.1, 0.], [ 0., 0.1]])
+E = np.array([[1,0], [0,1]])
+
+# Available control, possible disturbances
+U = input_bound *np.array([[-1., 1.], [-1., 1.]])
+W = uncertainty *np.array([[-1., 1.], [-1., 1.]])
+
+# Convert to polyhedral representation
+U = box2poly(U)
+W = box2poly(W)
+
+# Construct the LTI system describing the dynamics
+sys_dyn = hybrid.LtiSysDyn(A, B, E, None, U, W, cont_state_space)
+# @dynamics_section_end@
+
+# @partition_section@
+# Define atomic propositions for relevant regions of state space
+cont_props = {}
+cont_props['home'] = box2poly([[0., 1.], [0., 1.]])
+cont_props['lot'] = box2poly([[2., 3.], [1., 2.]])
+
+# Compute the proposition preserving partition of the continuous state space
+cont_partition = prop2part(cont_state_space, cont_props)
+plot_partition(cont_partition) if show else None
+# @partition_section_end@
+
+# @discretize_section@
+# Given dynamics & proposition-preserving partition, find feasible transitions
+disc_dynamics = discretize(
+    cont_partition, sys_dyn, closed_loop=False,
+    conservative=True,
+    N=5, min_cell_volume=0.1, plotit=show
+)
+# @discretize_section_end@
+
+# Visualize transitions in continuous domain (optional)
+plot_partition(disc_dynamics.ppp, disc_dynamics.ts,
+               disc_dynamics.ppp2ts) if show else None
+
+# Specifications
+# Environment variables and assumptions
+env_vars = {'park'}
+env_init = set()                # empty set
+env_prog = '!park'
+env_safe = set()                # empty set
+
+# System variables and requirements
+sys_vars = {'X0reach'}
+sys_init = {'X0reach'}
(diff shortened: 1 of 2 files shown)


johnyf:
Thanks, these commits will be rebased upon merging.

johnyf:
How would this change better be described in the changelog? (A draft can be found in a4f4cc330ca8fbad6061ee610d003667da595a39).

Why was the quadratic cost completely removed? Can it not be available whenever `cvxopt` is available, as commented here: https://github.com/tulip-control/tulip-control/issues/154#issuecomment-257048844? The dependencies will remain the same (`cvxopt` absent from `install_requires` in `setup.py`, and present in `requirements.txt`, as an optional dependency). If the 2-norm option can be kept, then should I simply copy the [old code](https://github.com/tulip-control/tulip-control/commit/52ba38a89fb128e29a4bd7e9eef002db50f8f345#diff-db73a44c4631fa5fc7c04bff0f7d729aL401) to the case `ord == 2`?

Why was the function renamed to [`get_input_helper_LP`](https://github.com/tulip-control/tulip-control/commit/52ba38a89fb128e29a4bd7e9eef002db50f8f345#diff-db73a44c4631fa5fc7c04bff0f7d729aR332)? It includes the case `ord == 2`, so it is not only linear programming (even though it raises an error for `ord == 2`). I support keeping the name `get_input_helper` (if renamed, following PEP8 it would be `get_input_helper_lp`).

The newly introduced keyword `ord` for the function `get_input_helper` precedes the existing keyword `closed_loop`. It will break existing callers. Is this ordering of keyword intentional? Does it based on relative importance of keyword arguments? (The latter is a good reason for placing it earlier.)

PEP8 [recommends](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments) ([here too](https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance)) to avoid abbreviations or corrupted spelling, when feasible. The name `ord` is short, but `order` is not too long, and self-documenting (to some degree). I would recommend using `order`, unless there is some specific reason for choosing `ord` (for example, uniformity with `numpy`'s API?)

There are two collections of commits above. Rebasing the later ones (with some conflict resolution) yields a patch clean of whitespace changes, which appears to differ from the rebase of the older ones only by 43e1e9b75b31afcec8fc989c2b1cf3c70854e72d, which I copied. I kept `cvxopt`, until item about `ord == 2` is resolved. It would help to confirm whether 52ba38a89fb128e29a4bd7e9eef002db50f8f345 is the intended result.

Main messages from `pep8 tulip/abstract/find_controller.py`:
- E231 missing whitespace after ','
- E101 indentation contains mixed spaces and tabs
- W191 indentation contains tabs
- E265 block comment should start with '# '
- E111 indentation is not a multiple of four

necozay:
Fine to put back the QP inside ``ord == 2``.  I would suggest one change for consistency though. I prefer ``R^TR`` and ``Q^TQ`` to be used in the QP to make it consistent with the other norms, i.e., ``||Rx||_ord``. This changes the meaning of ``R`` in earlier versions.

I agree with keeping the function name ``get_input_helper``.

I think `R`, `r`, `Q` and `ord` are all related to the cost function so it makes sense to have them together in the signature of `get_input_helper`. I suggest we make that change in the signature of `get_input` as well.

`ord` is used for uniformity with `numpy`'s API. I prefer to keep it. Otherwise, something like `norm_type` would be more informative.

johnyf:
I applied these changes as of 5e75e6dddfa3b2234f86cc472ab08e246109166a.
Placed `ord` after `R, r, Q` in args, `autopep8`, and some misc modifications to `find_controller`.

From what I understand (correct me if wrong) by a comparison, the QP:

```
old: x^T R x
new: |R x|^2 = (R x)^T (R x) = x^T (R^T R) x
```

So, to use the old code with the new input, I "symmetrize" the matrices `Q` and `R`, by using `Q2 = Q^T Q` and `R2 = R^T R` in place of `Q, R` as input to the old code. The addition of `mid_weight` is left untouched, and the same for all norms. Also, `0.5 r` is used, because `cvxopt` multiplies the quadratic term `x^T P x` with `0.5` and the docstring says `|Rx|_{ord} + |Qu|_{ord} + r'x + mid_weight * |xc - x(0)|_{ord}`.

Also, based on [this comment](https://github.com/tulip-control/tulip-control/issues/154#issuecomment-258577005), should we `assert closed_loop is False`?

The example `continuous_with_simulation.py` says: "WARNING:tulip.abstract.discretization:	 too small: si \ Pre(sj), so not reached it". Noted in case we should change something (probably not).

Also, except for the example, there are no tests for the new code. I think that when the implementation is completed (`closed_loop = True`) and the issues mentioned [here](https://github.com/tulip-control/tulip-control/issues/162#issuecomment-239633913) addressed, then tests can be added.

Note: This PR includes the simulation code from #146.

Besides the minor questions above, I consider this PR ready for merging.

johnyf:
Regarding tests, they are the subject of #166.

necozay:
QP interpretation is right. I was not aware of `cvxopt`s 1/2 factor but it makes sense.

I decided that it would make more sense to address ``closed_loop = True`` case separately. The current version is not any different than what ``closed_loop`` has been implementing up to now (modulo solver change but it should not affect feasibility). The code should still work as before so maybe no need to add an assertion. On the other hand, I think the latest changes in abstraction 0c8f8b8651b9d6e66be509ca3981893f2696eab7 might have introduced additional inconsistencies because I see ``closed_loop = True`` case failing more often. Again it is independent of this PRs goal. A warning about a potential bug might be useful though.

For the warning in the example `continuous_with_simulation.py`, see [this](https://github.com/tulip-control/polytope/issues/30) issue of polytope. If you try to run it 7-8 times, it should go through.

necozay:
If you have any thoughts on how to avoid "WARNING:tulip.abstract.discretization: too small: si \ Pre(sj), so not reached it" without updating `polytope` and enforcing a change in `polytope` version, let me know. Otherwise discretization fails sometimes for this example and maybe we should include a warning somewhere in the example docstrings until `polytope` is fixed

necozay:
I was trying to refer to the error message: `raise Exception("iterative_hull: "
Exception: iterative_hull: could not find starting simplex` in my two previous comments not the warning coming from `discretization`....

johnyf:
Thanks for the clarification, the example printed the warning a couple of times during abstraction, but it completed. I did not encounter this `Exception` in the example, but I will think about it.

johnyf:
Merged as 411aa2d9aeb886b1d74ca135f245ac268fa03d0e. A warning is now logged if `closed_loop = True` is given to `get_input`.

I agree for the implementation for the case `closed_loop = True` to be addressed outside the scope of this PR.

For posterity, the 1/2 factor in QP in `cvxopt` is mentioned [here](http://cvxopt.org/userguide/coneprog.html#quadratic-programming).

Given that the exception was not observed here, let's investigate it further in the relevant issue of `polytope`. I think that it is fine to make modifications there, given also that the next `polytope` release will be cut soon.

johnyf:
This PR makes `tulip` deployable in absence of `cvxopt`. This has been a long awaited objective, in particular:

- it completes the modifications outlined here: https://github.com/tulip-control/tulip-control/issues/48#issuecomment-41227876
- `cvxopt` was made optional in #122, but still required for `find_controller`. Now it is entirely optional, unless a quadratic cost is desired when designing a controller for the continuous state.

PR #177, opened 2017-02-26
Rm jtlv

### LINKED ISSUE #158: deprecate and eventually delete `interfaces.jtlv`
As far as I know, there are no current users of the JTLV-based solver interface, and no one wants to maintain it because the performance, features, etc. are available and better elsewhere. As such, we should deprecate and, later, delete `interfaces.jtlv` and any other uses of the JTLV-based solver in TuLiP.

Anyone interested in keeping it should comment here. Besides technical arguments about using the solver itself, other motivations can include the existence of legacy code and interesting use-cases that indirectly involve JTLV.

johnyf:
Closes #158.


CHANGES.md
```diff
@@ -1,6 +1,11 @@
 # tulip changelog
 
 
+## 1.3.1
```
slivingston: Should this be 1.4.0 ? Removal of the `jtlv` interface is not a backwards-compatible change.

There are several issues marked for [the version 1.3.1 milestone](https://github.com/tulip-control/tulip-control/milestone/8), but they can just as well be included in version 1.4.0.

slivingston: The relevant comment about version numbers from #158 is <https://github.com/tulip-control/tulip-control/issues/158#issuecomment-243085416>, which recommends version 1.4.0 upon removal.

slivingston:
Comments:

* Because some syntax parsing support for JTLV remains in [tulip/spec/translation.py](https://github.com/tulip-control/tulip-control/blob/9e597cff6324fc6c73cfe41bc32ba2a5234e564f/tulip/spec/translation.py), I recommend that a relevant URL or citation is provided somewhere in that module. E.g., we can reference <http://jtlv.ysaar.net/> in the module-level docstring at [line 36](https://github.com/tulip-control/tulip-control/blob/9e597cff6324fc6c73cfe41bc32ba2a5234e564f/tulip/spec/translation.py#L36). For posterity, note that the Internet Archive has a copy at <https://web.archive.org/web/20161025171415/http://jtlv.ysaar.net/>
* On [line 205 of tulip/spec/translation.py](https://github.com/tulip-control/tulip-control/blob/9e597cff6324fc6c73cfe41bc32ba2a5234e564f/tulip/spec/translation.py#L205), there is a reference to `interfaces.jtlv` in a docstring. It should be deleted. The text "Format is that of JTLV." can remain.
* There is a defunct comment on [lines 316-317 of translation.py](https://github.com/tulip-control/tulip-control/blob/9e597cff6324fc6c73cfe41bc32ba2a5234e564f/tulip/spec/translation.py#L316-L317) that refer to `interfaces.jtlv`.

slivingston:
All of your proposed changes are good. After my comments above are addressed, this will be ready for merging.

CHANGES.md
johnyf: I noticed that `jtlv` was marked with milestone 1.4.0, but thought that we may want to include it in 1.3.1 since that version was to be the next one. I agree about bumping to 1.4.0 to signify the API incompatibility.

johnyf:
The comments have been addressed. I amended edf4f5e but did not rebase any of the previous commits because that would change the commit hash mentioned in edf4f5e (and I've already repeated that cycle several times).

I bumped the version to 1.4.0 to ensure it agrees with the changelog and the absence of `interfaces.jtlv` from the tree.

A discussion about what syntax `jtlv` supports and whether this has been documented elsewhere apart from the parser grammar files that `jtlv` contains occurred in Aug 2014 in a [thread titled "jtlv syntax question"](https://sourceforge.net/p/tulip-control/mailman/message/32776812/) on the `tulip-discuss` mailing list.

PR #180, opened 2017-04-17
update to Python 3, while keeping Python 2.7 support

### LINKED ISSUE #44: python 3 compatibility
All the required dependencies work in Python 3. Ports `py33-numpy`, `py33-scipy`, `py33-networkx`, `py33-parsing` already exist in MacPorts. Although subport `py27-cvxopt` exists but no `py33-cvxopt` subport, as [noted](http://cvxopt.org/install/index.html) in `cvxopt`'s site it works in Python 3.x, so it should be mostly a matter of updating its macport.

Given the above, `tulip` can be ported to Python 3. A side-by-side comparison between Python 2 and 3 can be [read here](http://ptgmedia.pearsoncmg.com/imprint_downloads/informit/promotions/python/python2python3.pdf).

slivingston:
This pull request introduces changes such that TuLiP becomes primarily implemented in Python 3 and offers backwards-compatibility with Python 2.7 using [python-future](http://python-future.org/). 

This PR closes #44.

While working on these changes, I discovered that the test `transition_directions_test` in `abstract_test.py` can spuriously fail when `hash()` is salted. For example, using Python 2.7, try

    python -R run_tests.py -f abstract_test

The same test can spuriously fail in Python 3 without providing a special command-line switch to `python`. In `transition_directions_test`, a particular set of edges in the transition system are checked. Since the state numbers are checked against a fixed list, the test is not robust to permutations of these numbers. We should modify this test to allow isomorphic graphs or find a way to get predictable state numbers.


slivingston:
The future is upon us. I made the transition to Python 3 anew with incremental changesets that are each of narrow scope. The commit messages provide detail and references about motivations for changes, so I recommend reviewing this PR by simply reading commits sequentially. There are no new dependencies.

## Note on incidental changes

During my work, I found and fixed two errors that are not related to the goal of this PR per se. Notice commits that have prefix `BUG:`.

## Note on documentation, contrib/ and examples/

I did not audit API documentation nor the User's Guide. Because we are maintaining support for Python 2.7, the only critical concern is the use of deprecated syntax (e.g., `print` as a statement), which I addressed while also searching for the same in the source code.

I updated programs in the top-level of the directory examples/ and in the subdirectory gridworlds/. I also checked operation of examples/developer/fuel_tank/double_tank.py.

I made some changes under examples/developer/ and contrib/ as a consequence of `grep` and similar searches for affected code, but otherwise I did not extensively examine code in those places. My decision is consistent with our practice of best-effort but low priority support for code in those directories.

## Note on maintaining support for Python 2.7

Consistent with our discussion in #44, there are several cases where Python 3 is preferred with some performance penalty in Python 2.7 expected for large sizes of problems. This is motivated to avoid introducing a cross-version support layer, e.g., `python-future`, as a new dependency.

## Remaining concerns

Metaclasses are used in tulip/spec/ast.py. The syntax for metaclasses in Python 3 is not compatible with that of Python 2.7. This is announced in the "What's New" section of the Python documentation at <https://docs.python.org/3.5/whatsnew/3.0.html#changed-syntax>. We have at least two options:

1. avoid using metaclasses, or
2. create a compatibility layer.

For the latter, a good example is the function `with_metaclass` in [Python-Future](http://www.python-future.org). [The docstring of `with_metaclass`](https://github.com/PythonCharmers/python-future/blob/fe494df5d66db19af94a1fdc32afe0e2e52dcf66/src/past/utils/__init__.py#L24) describes the origin of the function as jinja2/_compat.py. (The same source code is in the Python-Future documentation: <http://www.python-future.org/_modules/future/utils.html#with_metaclass>)

johnyf:
@slivingston, thank you for the changes, especially for the transition to Python 3 without introduction of new dependencies. Only a few minor modifications were made.

- Why inline `import sys` in 03627ec4c5d09c4113309254f0d53dd9eb973899 ? The standard library module `sys` would not be expensive to import unconditionally.

- Chose the simplest option recommended, to remove use of metaclasses from the module `tulip.spec.ast`, because metaclasses are not used in any significant way by the AST.

- Addressed in 93898600e69b162aca7b7127630df229d39741f7 the error observed [above](https://github.com/tulip-control/tulip-control/pull/180#issue-222078370) by checking for graph isomorphism using the function [`networkx.is_isomorphic`](https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.algorithms.isomorphism.is_isomorphic.html), as done in [`yacc_test.py`](https://github.com/johnyf/promela/blob/0eb4cd925cfafc945f5b0a6e9886cb2170c0a49e/tests/yacc_test.py#L119) of the package `promela`.

- For posterity, using 0da52e2aee6ba71fb141d5a53a9573af8e4f2162, and `dd == 0.4.3` without CUDD (so the pure Python module `dd.bdd`), and `omega == 0.0.11`, the example `examples/developer/fule_tank/double_tank.py` takes 23 min.

PR #181, opened 2017-05-30
Add discrete version of bisimulation algorithm and dual-simulation algorithm

fyliu1204:
These two algorithms can be seen at http://web.eecs.umich.edu/~necmiye/pubs/WagenmakerO_allerton16.pdf.
Code list:
1. examples/developer/transys/transys_detailed_examples.py
Add bisimulation_reduced_example() and bisimulation_not_reduced_example() to show how to use the algorithm.
2. tulip/transys/transys.py
Add coarsest_possible_ppp(ts), list_to_fts(c_list, ts), bisimulation_algorithm(ts) and dual_simulation_algorithm(ts). These functions are the implementation of the two algorithms.

### DIFF
--- a/examples/developer/transys/transys_detailed_examples.py
+++ b/examples/developer/transys/transys_detailed_examples.py
@@ -408,7 +408,74 @@ def label_per_state():
     fts.states.add('s0', ap={'p'})
     fts.states.add('s1', ap=set())
     fts.plot()
+    
+def bisimulation_not_reduced_example():
+    ofts = trs.FiniteTransitionSystem()
+    ofts.states.add_from(['s1', 's2', 's3', 's4'] )
+    ofts.states.initial.add('s1')
+    ofts.atomic_propositions |= ['p1','p2']
+    apset = set()
+    apset.add('p1')
+    apset.add('p2')
+    ofts.states['s1']['ap'] = apset
+    ofts.states.add('s2', ap={'p1'} )
+    ofts.states.add('s3', ap={'p2'})
+    ofts.states.add('s4', ap={'p1', 'p2'})
+    ofts.transitions.add('s1', 's4') # unlabeled
+    ofts.sys_actions.add('try')
+    ofts.sys_actions.add_from({'start', 'stop'} )
+    ofts.env_actions.add_from({'block', 'wait'} )
+    ofts.transitions.add(
+        's2', 's1',
+        sys_actions='try', env_actions='block'
+    )
+    ofts.transitions.add(
+        's3', 's2',
+        sys_actions='start', env_actions='wait'
+    )
+    ofts.transitions.add(
+        's4', 's3',
+        sys_actions='stop', env_actions='block'
+    )
+    print ofts
+    fts1 = dual_simulation_algorithm(ofts)
+    print fts1
+    fts2 = bisimulation_algorithm(ofts)
+    print fts2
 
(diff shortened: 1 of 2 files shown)


johnyf:
Thanks for the PR. Some first comments:

- The [contributing recommendations](https://github.com/tulip-control/tulip-control/blob/5d69ce50b29b4baddfa537feec2acf3c61a3e92b/CONTRIBUTING.md) appear to have been taken into account overall. Issues to address:
    - Creating (unit) tests that are run as part of the test suite, instead of examples that aren't run automatically (so that Travis CI will run them). For example, in a file `tests/bisimulation_test.py`.
    - Docstring formatting conventions from [PEP 257](https://www.python.org/dev/peps/pep-0257/).
- The bisimulation algorithm relates to abstraction, so maybe it can be placed in a module under the subpackage `abstract`, for example `tulip.abstract.bisimulation`. This could provide a location for placing the implementation of other bisimulation algorithms too.
- The nesting of selection and iteration constructs within the main algorithm functions can probably be reduced by refactoring.
- Perhaps worth considering whether the implementation can be modified to use a `networkx.MultiDiGraph` directly (by annotating edges with keys-value pairs, using the graph directly for iterating over nodes, and some custom attributes for accompanying information, e.g., `g.initial_nodes`, `g.sys_actions`, ...) together with a small interface between a transition system instance given as input, and the internal implementation. Motivation for this suggestion is that it isolates the implementation from the interface to other modules. If there are API changes to `transys`, then it will suffice to update a 5-line interface at the entry and exit areas, instead of touching the algorithm's implementation. If replacing the methods used from `transys` makes the code verbose and unreadable, then such a change may not be very useful.

The Travis CI failure is irrelevant to this PR (clearly, given that files under `tests/` are unchanged). Unfortunately, I repeated the last Travis CI build, which used to pass, and [it failed](https://travis-ci.org/tulip-control/tulip-control/builds/219124310). The error appears to be due to [failure to fetch `lily`](https://travis-ci.org/tulip-control/tulip-control/builds/219124310#L2054), due to a security failure (`curl: (60) SSL certificate problem: unable to get local issuer certificate`), even though I can still download `lily`'s archive using my web browser.

johnyf:
a48a3913af226c75652ea228386576bc8b7ac07c addressed the `curl`-related error. More recent errors due to `numpy` have been addressed in `polytope == 0.2.0`.

@fyliu1204 any comments on the [above remarks](https://github.com/tulip-control/tulip-control/pull/181#issuecomment-305064462)?

slivingston:
@fyliu1204 (ping)

johnyf:
As suggested [above](https://github.com/tulip-control/tulip-control/pull/181#issuecomment-305064462) this addition can be moved to the subpackage `abstract`, which contains abstraction algorithms. I will relocate the code, before making other changes (unless suggested otherwise).

necozay:
As far as I remember, the implementation that is included in this PR is abstracting a finite transition system into another finite transition system. I think it is more appropriate if it is part of `transys` (just a function/method of `transys`). @zexiangliu is taking over this work to improve this PR and to also implement the version that goes from `hybrid` system classes to discrete transition systems.

johnyf:
Either location would work. We can point from the function docstring to the implementation that handles hybrid systems. I support preferring functions in module scope, except for cases this is clearly impractical.

PR #182, opened 2017-08-14
add method `GRSpec.declare`

### LINKED ISSUE #117: use `dict` to declare type
PEP 20: explicit is better than implicit. Inferring the type of a variable from the data type used to describe its domain is fragile. As of c8f821793035660b910e586cff625092193deb79:
- the value `boolean` signifies that the variable is of Boolean type
- a 2-`tuple` signifies that the variable is of integer type
- a `list` signifies that the variable is of enumerated type, where its possible values are restricted to strings (syntactically defined enclosing them in double quotes).

This approach mixes the type definition with the data type that one will use to represent the domain itself.
The most common error that I keep repeating is to pass an iterable of length 2 to define the limits of an integer domain, but not use a `tuple`. For example, if this is happening outside `tulip` and attention is focused on some other detail, then it is quite easy to not use a `tuple`, or perhaps use a `set` instead of a `list`.
The use of types above is to be avoided, for the same reasons that `isinstance` should be avoided, when possible.

The action item is to use two separate fields in the `dict` of a variable:
- a `"type"` key that takes `str` values, in particular in `{"bool", "int", "str"}`
  It may be tempting to admit multiple alternatives for each type, e.g., both "boolean" and "bool". By PEP 20: "There should be one-- and preferably only one --obvious way to do it." so this temptation is resisted here. The above type names have been borrowed from the implementation language (Python).
- a `"dom"` key that takes values:
  - for `bool` the value is ignored
  - for `int` the value is an iterable of length 2 that contains `int`
  - for `str` the value is an iterable of `str`

Note that this issue does not intend to suggest that a fix is necessary, perhaps rewriting from scratch some parts based on the experience gained so far will yield much better results.

johnyf:
Closes #117 (details are described in [this comment](https://github.com/tulip-control/tulip-control/issues/117#issuecomment-317257511)).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ TuLiP can be installed also `from PyPI <https://pypi.python.org/pypi/tulip>`_::
 
   pip install tulip
 
-This will install the required dependencies.
+This will install the latest release, together with required dependencies.
 To find out what dependencies (including optional ones) are installed, call::
 
   tulip.interfaces.print_env()
@@ -24,9 +24,12 @@ consult https://tulip-control.sourceforge.io/doc/install.html
 The next section describes how to build documentation.
 A test suite is provided under ``tests/``.  Consult the section "Testing" below.
 
-Pip can install the latest development version too::
+Pip can install the latest *development* snapshot too::
 
   pip install https://github.com/tulip-control/tulip-control/archive/master.zip
+
+Code under development can be unstable so trying `pip install tulip` first
+is recommended.
 
 
 Documentation
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,10 @@ consult https://tulip-control.sourceforge.io/doc/install.html
 The next section describes how to build documentation.
 A test suite is provided under ``tests/``.  Consult the section "Testing" below.
 
+Pip can install the latest development version too::
+
+  pip install https://github.com/tulip-control/tulip-control/archive/master.zip
+
 
 Documentation
 -------------
--- a/tests/form_test.py
+++ b/tests/form_test.py
@@ -114,7 +114,8 @@ class GRSpec_test(object):
                 self.f.sys_vars["y"] == "boolean")
 
 
-def test_declare():
+def test_declare_boolean_vars():
+    # declaring Boolean-valued variables
     g = GRSpec()
     g.declare('a')
     assert g.sys_vars == dict(a='boolean'), g.sys_vars
@@ -128,6 +129,10 @@ def test_declare():
     # attempt to redeclare "c" as sys var
     with nt.assert_raises(AssertionError):
         g.declare('c')
+
+
+def test_declare_int_vars():
+    # declaring integer-valued variables
     g = GRSpec()
     g.declare(i=[0, 10])
     assert g.sys_vars == dict(i=(0, 10)), g.sys_vars
@@ -138,6 +143,10 @@ def test_declare():
     # attempt to redeclare "i" as env var
     with nt.assert_raises(AssertionError):
         g.declare(i=(0, 10), env=True)
+
+
+def test_declare_str_vars():
+    # declaring string-valued variables
     g = GRSpec()
     # neither int nor str values
     with nt.assert_raises(TypeError):
--- a/tests/form_test.py
+++ b/tests/form_test.py
@@ -114,6 +114,40 @@ class GRSpec_test(object):
                 self.f.sys_vars["y"] == "boolean")
 
 
+def test_declare():
+    g = GRSpec()
+    g.declare('a')
+    assert g.sys_vars == dict(a='boolean'), g.sys_vars
+    assert g.env_vars == dict(), g.env_vars
+    g.declare('a', 'b')
+    assert g.sys_vars == dict(a='boolean', b='boolean'), g.sys_vars
+    assert g.env_vars == dict(), g.env_vars
+    g.declare('c', env=True)
+    assert g.sys_vars == dict(a='boolean', b='boolean'), g.sys_vars
+    assert g.env_vars == dict(c='boolean'), g.env_vars
+    # attempt to redeclare "c" as sys var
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -24,6 +24,10 @@ consult https://tulip-control.sourceforge.io/doc/install.html
 The next section describes how to build documentation.
 A test suite is provided under ``tests/``.  Consult the section "Testing" below.
 
+Pip can install the latest development version too::
+
+  pip install https://github.com/tulip-control/tulip-control/archive/master.zip
```
slivingston: This is a useful command to install directly from GitHub. However, there is some risk of confusion by new users who copy-and-paste this command without noticing or understanding the significance of "latest development version".

My recommendation is to keep this command but add clarifying text; e.g.,

1. earlier in the README, note that `pip install tulip` will get the latest "release" or "current official release"; or
2. add a brief blob of text suggesting that things can break without warning if the latest development version is used, and that most users should first try instead to use `pip install tulip`.

slivingston: also, this idiom may cause problem with version string identification. I obtain the following results on GNU/Linux Fedora 26 and Python 3.6.2. The shell prompt is `cero%`.
```sh
cero% virtualenv -p python3 f00f
Running virtualenv with interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/scott/tmp/f00f/bin/python3
Also creating executable in /home/scott/tmp/f00f/bin/python
Installing setuptools, pip, wheel...source done.
cero% source f00f/bin/activate
(f00f) cero% pip install https://github.com/tulip-control/tulip-control/archive/master.zip
Collecting https://github.com/tulip-control/tulip-control/archive/master.zip
  Downloading https://github.com/tulip-control/tulip-control/archive/master.zip
     \ 593kB 7.9MB/s
Collecting networkx<=1.10,>=1.8 (from tulip==1.4.0)
Collecting numpy>=1.7 (from tulip==1.4.0)
  Using cached numpy-1.13.1-cp36-cp36m-manylinux1_x86_64.whl
Collecting omega<0.1.0,>=0.0.9 (from tulip==1.4.0)
Collecting ply>=3.4 (from tulip==1.4.0)
Collecting polytope>=0.2.0 (from tulip==1.4.0)
Collecting pydot>=1.2.0 (from tulip==1.4.0)
Collecting scipy (from tulip==1.4.0)
  Using cached scipy-0.19.1-cp36-cp36m-manylinux1_x86_64.whl
Collecting decorator>=3.4.0 (from networkx<=1.10,>=1.8->tulip==1.4.0)
  Using cached decorator-4.1.2-py2.py3-none-any.whl
Collecting dd==0.4.3 (from omega<0.1.0,>=0.0.9->tulip==1.4.0)
Collecting natsort>=3.5.3 (from omega<0.1.0,>=0.0.9->tulip==1.4.0)
  Using cached natsort-5.1.0-py2.py3-none-any.whl
Collecting pyparsing>=2.1.4 (from pydot>=1.2.0->tulip==1.4.0)
  Using cached pyparsing-2.2.0-py2.py3-none-any.whl
Collecting psutil>=3.2.2 (from dd==0.4.3->omega<0.1.0,>=0.0.9->tulip==1.4.0)
Collecting astutils>=0.0.1 (from dd==0.4.3->omega<0.1.0,>=0.0.9->tulip==1.4.0)
Requirement already satisfied: setuptools>=19.6 in ./f00f/lib/python3.6/site-packages (from dd==0.4.3->omega<0.1.0,>=0.0.9->tulip==1.4.0)
Installing collected packages: decorator, networkx, numpy, ply, pyparsing, pydot, psutil, astutils, dd, natsort, omega, scipy, polytope, tulip
  Running setup.py install for tulip ... done
Successfully installed astutils-0.0.2 dd-0.4.3 decorator-4.1.2 natsort-5.1.0 networkx-1.10 numpy-1.13.1 omega-0.0.11 ply-3.10 polytope-0.2.0 psutil-5.2.2 pydot-1.2.3 pyparsing-2.2.0 scipy-0.19.1 tulip-1.4.0
(f00f) cero% python -c 'import tulip; print(tulip.__version__)'
1.4.0
```

tests/form_test.py
```diff
@@ -114,6 +114,40 @@ def test_init(self):
...
+    g.declare('a', 'b')
+    assert g.sys_vars == dict(a='boolean', b='boolean'), g.sys_vars
+    assert g.env_vars == dict(), g.env_vars
+    g.declare('c', env=True)
+    assert g.sys_vars == dict(a='boolean', b='boolean'), g.sys_vars
+    assert g.env_vars == dict(c='boolean'), g.env_vars
+    # attempt to redeclare "c" as sys var
+    with nt.assert_raises(AssertionError):
+        g.declare('c')
+    g = GRSpec()
```
slivingston: This function is difficult to read because it lacks blank lines to separate distinct (in terms of state) parts. I recommend to either insert blank lines before `g = GRSpec()` here and on line 141, or to move them into separate test functions.

slivingston: in particular, I am thinking of the use-case where `test_declare` fails and an assertion error is presented with the corresponding line. How quickly can the program state at the point of error be determined?

README.rst
johnyf: Thanks for the observations, I applied the comments. Version confusion is indeed an issue, but the result is identical with using the first command listed in the README (`pip install .`). Both yield a version of 1.4.0. The latter will result in a development version identifier only if `gitpython` is present (in which case the user is aware of development dependencies, or use `gitpython` in their own work).

tests/form_test.py
johnyf: I split the test into multiple functions, per type of values that the declared variables take. This should make it faster to track the program's state up to the point where an assertion has failed.

johnyf:
I created new commits to keep the pull request history linear, thus avoiding GitHub marking changes as "outdated". I am fine with either merging by rebase to squash these commits on those they amend, or merging directly.

PR #190, opened 2018-01-29
Treat new form of non-release gr1c version strings

slivingston:
Since [commit cdc3ec7059ee0a2bf021394d0ac8b7a37a29df0c](https://github.com/tulip-control/gr1c/commit/cdc3ec7059ee0a2bf021394d0ac8b7a37a29df0c), gr1c builds that do not correspond to releases have version strings with "-devel-" and, if available, text of `git describe`.

E.g., with a sourctree that has pending changes, the first line of `gr1c -V` can be

    gr1c 0.13.0-devel-v0.12.0-5-ge633890-dirty

which distutils.version.StrictVersion() does not accept.

In this pull request, I add a conditional branch to tulip.interfaces.gr1c.check_gr1c() that is taken if `'devel'` is found. In it, everything after and including the first "-" is stripped.

### DIFF
--- a/tulip/interfaces/gr1c.py
+++ b/tulip/interfaces/gr1c.py
@@ -43,7 +43,7 @@ Use the logging module to throttle verbosity.
 """
 from __future__ import print_function
 
-from distutils.version import StrictVersion
+from pkg_resources import parse_version
 import logging
 import copy
 import os
@@ -69,7 +69,7 @@ def check_gr1c():
     except OSError:
         return False
     v = v.split()[1]
-    if StrictVersion(v) >= StrictVersion(GR1C_MIN_VERSION):
+    if parse_version(v) >= parse_version(GR1C_MIN_VERSION):
         return True
     return False
 


johnyf:
I tested this change with a `gr1c` executable that returns `gr1c 0.13.0-devel-v0.12.0-8-g494dc1f` as version specifier. The function `check_gr1c` returned `True`, as expected.

I think this change is ready to be merged.

I tested with a modified `gr1c` that returns `gr1c 0.13.0-foo-v0.12.0-8-g494dc1f-dirty` as version specifier (after changing [this line](https://github.com/tulip-control/gr1c/blob/494dc1fb7678e123d6ba4fd8dbc66c62dffde050/Makefile#L48) of the `Makefile`), and the result is a `ValueError`. This case is outside the possible specifiers, but may become relevant if the string "develop" changes in the future. I mention it for completeness.

In the tests we use `setuptools.version.pkg_resources.parse_version`:

https://github.com/tulip-control/tulip-control/blob/e3de77bc55cef15023e49363787b6c90045819c1/tests/version_test.py#L66

One possible approach for parsing the version specifier of `gr1c` is to preprocess the specifier and call `parse_version`, for example:

```python
version = 'gr1c 0.13.0-devel-v0.12.0-8-g494dc1f'
s = version.lstrip('gr1c ')
s.replace('-devel-', '.dev0+')
v = pkg_resources.parse_version(s)
```

which returns a `Version` (not a `LegacyVersion`). The additional information wouldn't be used in this case though. Worth noting that `distutils.version.StrictVersion` fails to parse the specifier even after this preprocessing.

Also, looking back at when each line was introduced:

https://github.com/tulip-control/tulip-control/commit/9cbfd3caf04a03aea300855a899aec66c5dcf5a2#diff-db20ea5eeef1f4a2a77da94c72498198R44

https://github.com/tulip-control/tulip-control/commit/110adcb95051b79f343b0c98d7807c58b6529ad7#diff-aa1c501e0152c493a4d509b8de00d6f0R9

it seems that `pkg_resources.parse_version` was selected after `distutils.version.StrictVersion`, so probably the choice of `distutils` was not for any specific reason. I support using `parse_version` here too, because it implements [PEP 440](https://www.python.org/dev/peps/pep-0440/), which [supersedes](https://www.python.org/dev/peps/pep-0440/#version-ordering-across-different-metadata-versions) [PEP 386](https://www.python.org/dev/peps/pep-0386/).

slivingston:
I can as well change gr1c to follow PEP 440, so the first example that you gave would instead be

    gr1c 0.13.0.dev0+v0.12.0-8-g494dc1f

This solution allows us to avoid string replacement here, and it would simplify use of gr1c in other contexts where PEP 440 is accepted.

If you agree, then I will rebase with a new commit that uses `pkg_resources.parse_version`. Concurrently, I will provide a new commit to the gr1c repo to implement the PEP 440 there.

johnyf:
I agree with implementing PEP 440, and rebasing this PR before merging (or alternatively at the time of merging).

johnyf:
I tested the changes with a `gr1c` executable that returns `gr1c 0.13.0.dev0+v0.12.0-9-g17a382c` as version specifier, and with an earlier version that returns a specifier as [above](https://github.com/tulip-control/tulip-control/pull/190#issuecomment-361448355). The results from `check_gr1c` are as expected. This PR can be merged.

PR #193, opened 2018-02-17
bi/dual-simulation algorithms for FTS

zexiangliu:
Alternative implementation for the algorithms  https://github.com/tulip-control/tulip-control/pull/181

Code: 
1. Add function ``simu_abstract`` in ``tulip.transys.transys`` package
```python
def simu_abstract(ts,simu_type)
# ts: FTS,  simu_type: 'bi' or 'dual'
```
2.  Test code ``transys_simu_abstract_test.py`` is added in ``tests/``, and registered in ``run_tests.py``.

Notes:
- The algorithms are implemented based on ``networkx.MultiDiGraph``, with interfaces to ``FTS`` at the beginning and end of ``simu_abstract``. 
- A queue structure is used to reduce required iterations in main algorithm. 
- The output FTS doesn't contain action (edge) attributes, in order to avoid messy implementation.
- Since the algorithms are only for ``FTS`` class, the function is placed under `` tulip.transys.transys ``.
- All base tests are passed in my computer. Full tests may fail (``version_test`` fails since no module ``__version`` is found).


tulip/transys/transys.py
```diff
@@ -786,3 +789,165 @@ def __init__(self):
...
+    @return: the bi/dual simulation, and the corresponding partition.
+    @rtype: L{FTS}, C{dict}
+    
+    References
+    ==========
+    1. Wagenmaker, A. J.; Ozay, N.
+    A Bisimulation-like Algorithm for Abstracting Control Systems. 
+    54th Annual Allerton Conference on CCC 2016
+    """
+    def pre(graph,list_n):
```
johnyf: I support moving the function `pre` outside of the function `simu_abstract`, and hiding it by renaming to `_pre`. The definition of `pre` is independent of the arguments of `simu_abstract`, so this change is possible.

tulip/transys/transys.py
```diff
@@ -786,3 +789,165 @@ def __init__(self):
...
+        @return: set of predecessors of C{list_n}
+        @rtype: C{set}
+        """
+        # The simplest pre is implemented. Modify this fun to ContPre if
+        #  necessary.
+        pre_set = set()
+        for n in list_n:
+            pre_set = pre_set.union(graph.predecessors(n))
+        return pre_set
+    def output_FTS(ts,Part):
```
johnyf: Similarly to `pre`, the function `output_FTS` can be moved outside of `simu_abstract`, and renamed to `_output_fts`, in order to hide it and following PEP 8 ([snake_case](https://en.wikipedia.org/wiki/Snake_case)).

tulip/transys/transys.py
```diff
@@ -786,3 +789,165 @@ def __init__(self):
...
+                num_cell += 1
+                if(i==j):
+                    break
+#        print 'the queue size is', queue.qsize()
+#        print 'num of cell is', num_cell
+                    
+    # construct new FTS
+    [ts_simu, part_hash] =output_FTS(ts,Part)
+    
+    return ts_simu, part_hash
```
johnyf: A newline needs to be added at the end of each file.

johnyf:
Thanks for these additions. We can squash the commits at the time of merging, since this is fresh code.

## CUDD website unavailable

Tests on Travis CI pass except for those after installing `dd.cudd`, which fails due to ftp://vlsi.colorado.edu/pub/cudd-3.0.0.tar.gz being unavailable (https://travis-ci.org/tulip-control/tulip-control/jobs/342809972#L1443) (the website from where CUDD is available is unreachable, also from my machine). Failure to fetch CUDD leads to absence of `dd` when running the next round of tests, so failure of `omega` to import (because it depends on `dd` being present, even without `dd.cudd`, but `dd` installation halts upon failing to fetch CUDD).

We could consider fetching CUDD from a GitHub repository (e.g., https://github.com/johnyf/cudd). This is the first time over 6 years that I have encountered an issue with downloading CUDD.


## `version_test`

Regarding `version_test` failing due to absence of the module `tulip._version`:

https://github.com/tulip-control/tulip-control/blob/f704eaec18335a9bdbb4ad279c854756a10090e6/tests/version_test.py#L12

The module `_version` is auto-generated by `setup.py`. So if `setup.py` is not run at least once after cloning the repository, then `_version` is absent. One may prefer `pip install .` (or `-e`) instead of running `setup.py`. So this is a good point that running `version_test` requires first running `setup.py`. Up to now we happened to be running `setup.py` on Travis CI:

https://github.com/tulip-control/tulip-control/blob/f704eaec18335a9bdbb4ad279c854756a10090e6/.travis.yml#L29

but should either note that `setup.py` needs to be run before `version_test.py`, or have `version_test.py` itself run `setup.py` (though this is more complicated).

(Also, note the single underscore in `tulip._version`. A double underscore in Python leads to name mangling in some cases, whereas a single underscore by convention signifies hiding.)

zexiangliu:
@johnyf  Thanks for the comments! The problems mentioned above are fixed in the latest commit.

For the ``_version`` problem, I run the ``setup.py`` first, but it still gets that error. I notice that ``_version.py`` is created under the directory ``.local/lib/python2.7/site-packages/tulip/`` for Ubuntu. I guess ``run_tests.py`` automatically search packages under the root directory of source code but not the site-package one.

tulip/transys/transys.py
```diff
@@ -37,6 +37,9 @@
 from tulip.transys.labeled_graphs import (
     LabeledDiGraph, str2singleton, prepend_with)
 from tulip.transys.mathset import PowerSet, MathSet
+from networkx import MultiDiGraph
+from Queue import Queue
```
johnyf: Thanks for the updates. The Python 2 module [`Queue`](https://docs.python.org/2/library/queue.html) has been renamed to [`queue`](https://docs.python.org/3/library/queue.html) in Python 3, which causes some of the tests to fail on Travis CI. A simple solution would be:

```python
try:
    from Queue import Queue
except ImportError:
    from queue import Queue
```

johnyf:
I think that documenting the version test can be addressed outside the scope of this PR (I mentioned it above for completeness). If installing with `pip install .`, then `tulip/_version.py` is not created. With f704eaec18335a9bdbb4ad279c854756a10090e6, running `python setup.py install` creates `_version.py` under `tulip/` within the repository.

The location from where tests import `tulip` can be selected with `run_tests.py --outofsource`.

tulip/transys/transys.py
```diff
@@ -786,3 +789,172 @@ def __init__(self):
...
+                 'values': ts.sys_actions,
+                 'setter': True}]
+    ts_simu = FTS(env_actions,sys_actions)
+    ts2simu = {}
+    simu2ts = {}
+    Part_hash = {'ts2simu':ts2simu,'simu2ts':simu2ts}
+    for i in Part:
+        simu2ts[i] = Part.node[i]['cov']
+        for j in Part.node[i]['cov']:
+            if(ts2simu.has_key(j)):
```
johnyf: The method `dict.has_key` has been removed in Python 3.

johnyf:
I replaced fetching of CUDD sources with a GitHub URL (https://github.com/johnyf/cudd/archive/v3.0.0.tar.gz) to address the broken link. Except for some CRLF in the official distribution of CUDD v3.0.0 (ftp://vlsi.colorado.edu/pub/cudd-3.0.0.tar.gz), the GitHub archive has the same contents. In other words:

```
$ diff v3.0.0.tar.gz cudd-3.0.0.tar.gz
Binary files v3.0.0.tar.gz and cudd-3.0.0.tar.gz differ
```

But after extracting the contents, `diff -r --strip-trailing-cr` shows no file differences (some files under `cudd-3.0.0/dddmp/` contain CRLF).

This approach is used for `dd == 0.4.3`. For `dd >= 0.5.3` this approach should be unnecessary, because wheel files containing CUDD are available from PyPI.

These changes are on branch `pr_193`. The tests on Travis CI pass [1].

[1] https://travis-ci.org/tulip-control/tulip-control/builds/344027822

johnyf:
@zexiangliu the branch `pr_193` includes changes for Python 3.

johnyf:
Other comments:

- parentheses are unnecessary around expressions in a conditional
- trailing whitespace was present
- `autopep8` can be useful for formatting
- `assert` is a statement, not a function, so parentheses can be omitted
- test functions can return `None` (the default in absence of a `return` statement).

(I have applied these changes.)

johnyf:
I have applied some further changes on branch `pr_193`, including downloading CUDD from: https://sourceforge.net/projects/cudd-mirror/files/. I think that this PR can be merged. If there are no other comments, I will merge branch `pr_193` into branch `master` on Feb 26 (by squashing the commits that introduce new code).

zexiangliu:
I detected one logic fault in the dual algorithm yesterday, and our test didn't tell us that at all. So right now I am still working on this and also try to design a more "powerful" test. I will try to make that done before Feb. 26th.

johnyf:
It would be convenient if further changes were applied on top of branch `pr_193`, e.g., by pulling that branch, applying changes, and `git push -f` to the branch that this PR is based on.

johnyf:
A few suggestions that may be useful in case testing is under development:

- adding assertions within the code, for example at entry and exit of functions, or checking invariants within the loops, and
- refactoring the loops to extract conditional statements as separate small functions, which can make it easier to reason about what assertions to check.

tulip/transys/transys.py
```diff
@@ -960,7 +960,11 @@ def simu_abstract(ts, simu_type):
                     lookup[num_cell] = True
                     num_cell += 1
                     if i == j:
+                        Part.remove_edge(num_cell-1, num_cell-1)
                         break
+            if queue.qsize()>1000:
```
necozay: i kind of don't like hard-coded numbers in the middle of the code. is this part with 1000 crucial?

zexiangliu: it is part of debuging code, which will be removed later.

zexiangliu:
- add one more testing, whose FTS is generated from a continuous system
- rewrite `simu_abstract`, whose structure will be consistent with function `discretize` under `abtract`, and pass all the test

Note: the new test for `simu_abstract` costs more time than the others, due to the need of generating FTS from continuous system.

johnyf:
The download location ftp://vlsi.colorado.edu/pub/cudd-3.0.0.tar.gz is again reachable.

johnyf:
The download location ftp://vlsi.colorado.edu/pub/cudd-3.0.0.tar.gz is again unreachable.

johnyf:
Merged after rebasing as of 9d4f335e89f296ff4f583b7b3118a9c5329887d5. Changes:

- `transys_simu_abstract_test.py` takes about 14 sec (with `cvxopt`) or 3min (with `scipy` and not `cvxopt`) on my machine, so I set `simu_abstract_test.slow = True`.
- moved the tests to the family `hybrid`: The addition is to `transys`, which is independent of `polytope`. The tests though use `polytope`, `tulip.abstract`, `tulip.hybrid`, so they have been moved to the test family `hybrid`. See also #196.
- unit tests would be useful, i.e., very simple tests that use only `transys` and `networkx` (those in `transys_simu_abstract_test.py` could be regarded as integration tests).
- Other changes were mainly formatting.

Issue #218, opened 2019-10-30
add support for Python 3.7 and Python 3.8; remove support for Python 3.4


slivingston:
this issue is entirely similar to https://github.com/tulip-control/dd/issues/46

in summary, 3.7 and 3.8 are current versions of Python, whereas 3.4 is past end-of-life.

slivingston:
done in work from recent PRs, now merged into `master` branch.

slivingston:
Note that tests continue to pass on Python 3.4, so it remains [in the list of Travis CI environments](https://github.com/tulip-control/tulip-control/blob/bac6f7811d1077400c0d16b767b631a30b7db8ff/.travis.yml#L3-L4) until there is otherwise motivation to remove it.

slivingston:
As part of the work on pull request #237 to add stormpy in CI jobs, we switched CI jobs to be on Ubuntu 18.04 ("Bionic"). However, [Travis does not support Python 3.4 on it](https://travis-ci.community/t/python-3-4-not-found-on-bionic/7034), so I am going to remove Python 3.4 from .travis.yml.

PR #220, opened 2020-01-22
Update dependencies and fix compatibility with python3

tichakorn832:
WARNING: THIS IS NOT TO BE MERGED YET. I'm opening a PR to see whether I'm on the right track.

The purpose of this PR is to fix compatibility with python3 (the target is python 3.7) and update requirements and the setup script to the more up-to-date versions. During this change, many tests fail, which I'm trying to fix.

* dumpsmach_test now runs with python 3.7, provided that a minor fix to the omega package is made.

FYI: @slivingston @murrayrm

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,10 +96,10 @@ def run_setup():
         license='BSD',
         classifiers=classifiers,
         install_requires=[
-            'networkx >= 1.8, <= 1.10',
+            'networkx >= 2.0, <= 2.4',
             'numpy >= 1.7',
-            'omega >= 0.0.9, < 0.1.0',
-            'ply >= 3.4',
+            'omega >= 0.1.2',
+            'ply >= 3.4, <= 3.10',
             'polytope >= 0.2.0',
             'pydot >= 1.2.0',
             'scipy'],
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,7 @@ scipy
 # easy extras
 omega==0.1.2
 gr1py>=0.1.0
+psutil==5.6.7
 
 # extras (uncomment as needed)
 # cvxopt==1.1.7  # for faster polytopic operations
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,12 @@
 # usage: `pip install -r requirements.txt`
 pip>=7.1.2
 
-ply==3.4
-networkx==1.6
-numpy==1.7
+ply==3.10
+networkx==2.4
 scipy
 
 # easy extras
+omega==0.1.2
 gr1py>=0.1.0
 
 # extras (uncomment as needed)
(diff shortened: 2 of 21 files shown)


murrayrm:
Is there a reason to generate the PR now rather than just testing things out in the branch and then doing the PR once things are fixed?  I don't think it really matters, but this PR will get a bit cluttered as you debug...

tichakorn832:
I created a PR early to (hopefully) get some help to make sure that I'm on the right track. (I agree that a branch would work too but with a PR, it's easier to provide comments.)

slivingston:
for future reference, it is possible to mark a pull request as a [draft](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests).

@tichakorn832 is there any reason to not also support Python 3.8 ?

slivingston:
also, the issue about supporting Python 3.7 and 3.8: #218

tichakorn832:
Oh I didn't know about draft. Thanks for pointing out.

I just haven't tested on python3.8. There is no change that is specific to 3.7 though.

slivingston:
@tichakorn832 let me know if I can be useful here. We are not using it in Tulip yet, but a good tool for testing against a variety of Python versions is [tox](https://tox.readthedocs.io/en/latest/) ([on PyPI](https://pypi.org/project/tox/)).

tichakorn832:
@slivingston Thanks. it would be nice if you could help take a look at the change in tulip/interfaces/omega.py to make sure that I formulate it correctly with respect to the new interface in omega. There seems to be interface change in the latest version of omega, making the previous code in tulip not working anymore. All the tests run fine except test_spec_realizable_bool  in gridworld_test.py that keeps running out of memory.

slivingston:
I am able to locally run test_spec_realizable_bool without error on GNU/Linux with Python 3.7.6.

@tichakorn832 Are you able to run the test locally? If so, perhaps the error arises from the configuration of Travis CI hosts. I will investigate.

setup.py
```diff
@@ -96,10 +96,10 @@ def run_setup():
         license='BSD',
         classifiers=classifiers,
         install_requires=[
-            'networkx >= 1.8, <= 1.10',
+            'networkx >= 2.0, <= 2.4',
             'numpy >= 1.7',
-            'omega >= 0.0.9, < 0.1.0',
-            'ply >= 3.4',
+            'omega >= 0.1.2',
```
johnyf: I suggest pinning to `omega == 0.2.0` or in a range `omega >= 0.1.2, <= 0.2.0`, to allow for changes in the API of `omega`.

slivingston: @johnyf I agree. I am preparing several commits to add to this PR from my temporary branch `tmp-debug-travisci`.

murrayrm: Planning to accept this change as is, and let @slivingston fix this in `tmp-debug-travisci`.

slivingston: Confirmed. I will open a separate PR with my changes and those from the branch `completed_update_to_networkx_2` by @johnyf

PR #224, opened 2020-02-18
Jupyter notebook examples

murrayrm:
This PR adds some Jupyter notebooks examples and updates the `gridword`  animation functionality to be more  compatible with inline animations.


murrayrm:
Forgot that this includes changes from `slivingston/tmp-debug-travisci`.  Need to rebase on top  of master.

murrayrm:
@slivingston Would be great if you could do a quick review of this since I updated `tulip/gridworld.py` to handle animations a bit differently.

slivingston:
Both Jupyter notebooks work as expected.

tulip/gridworld.py
```diff
@@ -1256,13 +1258,13 @@ def update_line(num, dlist, lines):
         lines.append((l, l_trail))
 
     if not save_prefix:
-        ani = matplotlib.animation.FuncAnimation(fig, update_line,
-                                                 len(paths[0]),
-                                                 fargs=(data, lines),
-                                                 interval=500)
-        plt.show()
```
slivingston: Deleting `plt.show()` here is significant because, after this change, the user must call `matplotlib.pyplot.show()` (or similar) to display the animation.

I think this is a good change, but I will add a comment to the docstring to help users who might expect the animation to immediately appear.

PR #227, opened 2020-02-24
`omega` interface update and related changes

johnyf:
These changes:

- in the module `tulip.interfaces.omega`:
  - remove unused imports of `dd` and `dd.cudd`
  - remove the unused argument `use_cudd`
- add `dd` and `dd.cudd` information to the function `tulip.interfaces._inspect.print_env`
- run the test `gridworld_test.GridWorld_test.test_spec_realizable_bool` if `dd.cudd` is present
- update the docstring of the method `tulip.transys.labeled_graphs.LabeledDiGraph.add_edge`
- in `install_requires` require the version range `omega >= 0.2.0, < 0.3.0`.

### DIFF
--- a/tests/gridworld_test.py
+++ b/tests/gridworld_test.py
@@ -9,7 +9,6 @@ except ImportError:
 import numpy as np
 import tulip.gridworld as gw
 import unittest
-import psutil
 
 from tulip.synth import is_realizable
 
--- a/tests/gridworld_test.py
+++ b/tests/gridworld_test.py
@@ -2,6 +2,10 @@
 from __future__ import division
 from __future__ import print_function
 
+try:
+    from dd import cudd as dd_cudd
+except ImportError:
+    dd_cudd = None
 import numpy as np
 import tulip.gridworld as gw
 import unittest
@@ -150,8 +154,8 @@ class GridWorld_test(object):
     def test_dumploadloop(self):
         assert self.X == gw.GridWorld(self.X.dumps())
 
-    @unittest.skipIf(psutil.virtual_memory().total < 64e9,
-                     "not enough memory available")
+    @unittest.skipIf(dd_cudd is None,
+                     '`dd.cudd` not installed')
     def test_spec_realizable_bool(self):
         spec = self.X.spec(nonbool=False)
         spec.moore = False
--- a/tulip/interfaces/_inspect.py
+++ b/tulip/interfaces/_inspect.py
@@ -13,6 +13,14 @@ try:
 except ImportError:
     cvxopt_glpk = None
 try:
+    import dd
+except ImportError:
+    dd = None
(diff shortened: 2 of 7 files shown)


slivingston:
I will review this today.

tests/gridworld_test.py
```diff
@@ -150,8 +154,8 @@ def test_equality(self):
     def test_dumploadloop(self):
         assert self.X == gw.GridWorld(self.X.dumps())
 
-    @unittest.skipIf(psutil.virtual_memory().total < 64e9,
```
slivingston: Following this change, `import psutil` can be removed from the file.

johnyf: I could rebase to remove the `import psutil`.

slivingston: I think it is enough to add another commit, no rebase, unless you prefer rebase.

tulip/interfaces/_inspect.py
```diff
@@ -49,10 +57,19 @@ def print_env():
         'gr1py', gr1py_int.gr1py, 'https://pypi.python.org/pypi/gr1py')
     c.append(s)
     s = _format_python_package_message(
-        'omega', omega_int.omega, 'https://pypi.python.org/pypi/omega')
+        'dd', dd, 'https://pypi.python.org/pypi/dd')
+    c.append(s)
+    if dd_cudd is None:
+        s = (
+            'Could not import Cython module `dd.cudd`.\n'
```
slivingston: This is a good change because it is more useful feedback to users.

tests/gridworld_test.py
slivingston: Then, we can also remove `psutil` from requirements.txt.

slivingston: It was introduced in commit 0c1b77ce5355f272a2ddece2f8bccdb42f479f35.

To be clear, I am not opposed to using `psutil`. We might have reason to depend on it again in the future. This change is simply to keep the list of dependencies up-to-date.

tests/gridworld_test.py
```diff
@@ -150,8 +154,8 @@ def test_equality(self):
...
-                     "not enough memory available")
+    @unittest.skipIf(dd_cudd is None,
```
slivingston: Excellent. This is a better solution because it allows the test to be performed on memory-constrained providers such as Travis CI.

slivingston:
@johnyf Thanks! This is ready to merge after we add a commit that:
1. removes `import psutil` from gridworld_test.py
2. removes `psutil` from requirements.txt

slivingston:
I can do it, and merge, if you are busy.

johnyf:
Thank you for reviewing @slivingston ! I added a commit with these changes.

Issue #228, opened 2020-02-26
examples/continuous_with_simulation.py not working


murrayrm:
I’m trying to run a variant of `examples/continuous_with_simulation.py` (reformatted as a Jupyter notebook, but same basic code).  Jupyter notebook version attached.
[continuous_with_simulation.ipynb.txt](https://github.com/tulip-control/tulip-control/files/4257116/continuous_with_simulation.ipynb.txt)

* Everything works up until the simulation, though I am not sure I believe the discrete abstraction (it looks like there are some transitions across “corners”; is that plausible?).

* When I try to run the simulation at the end, I get a situation where the computed trajectory does not take me to the desired discrete state.  I don’t get an error, but it looks like I am right on the edge of the new partition and so I get “stuck” in the wrong discrete state.

* The controller is “open loop” in the sense that it uses its own internal discrete state assuming that the continuous system is a proper simulation.  But because that wasn’t actually the case, we end up getting out of sync…

It seems that there are other related issues (#146, #170, #209).

Issue #56, opened 2014-01-30
Merging original partitions of pwa modes


johnyf:
As of ef252bdd49641d3b5f6253cbec1c39b5dc99026f `abstract.discretization.AbstractSysDyn.original_regions` stores a list of `Region` objects of the `PropPreservingPartition` from which the discretization started.

But `discretize` [firstly partitions](https://github.com/tulip-control/tulip-control/blob/ef252bdd49641d3b5f6253cbec1c39b5dc99026f/tulip/abstract/discretization.py#L204) the given `PropPreservingPartition` based on the domains of subsystems in a `PwaSubSys`, then discretizes. Consider the case of several modes, each mode a `PwaSysDyn`. If the domains of subsystems differs between the modes, then even if the same partition is given as initial for each mode, `discretize` will further cut it into pieces differently for each mode.

For this reason for each mode, the [resulting](https://github.com/tulip-control/tulip-control/blob/ef252bdd49641d3b5f6253cbec1c39b5dc99026f/tulip/abstract/discretization.py#L546) `AbstractSysDyn.original_regions` will have different original `Region`s than the other modes.

The original regions are important in non-conservative reachability analysis. Therefore they relate closely to the dynamics and may better be different for each mode. This is an argument against:
- using for each mode `pwa_partition` prior to `discretize`
- merge these `PropPreservingPartitions`
- then pass each as seed to `discretize` and tell `discretize` to skip calling `pwa_partition`

Upon merging, feasibility of transitions between cells of the merged partition [is checked using](https://github.com/tulip-control/tulip-control/blob/ef252bdd49641d3b5f6253cbec1c39b5dc99026f/examples/developer/fuel_tank/tank_functions.py#L135) the original regions for that mode (if non-conservative check). This seems as the appropriate way, to avoid imposing unnecessary conservatism.

The issue is that these `original_regions` differ by mode, so they cannot be just one list [when stored](https://github.com/tulip-control/tulip-control/blob/ef252bdd49641d3b5f6253cbec1c39b5dc99026f/examples/developer/fuel_tank/tank_functions.py#L247) in `AbstractSysDyn`. They have to become a `dict` keyed by mode. This enables using them later for Receding Horizon Control or Multi-Parametric Programming.

In order to support this, either:
- `AbstractSysDyn` must allow `original_regions` to be a `dict`, or
- a new class for switched abstractions be created.

Not sure what is cleaner, more elegant and ergonomic.
[...]

PR #230, opened 2020-02-28
Fix continuous with simulation example

murrayrm:
This PR, combined with [polytope/PR #56](https://github.com/tulip-control/polytope/pull/56), fixes issue #228 and a couple of other small bugs/enhancements:

* There was a small bug in `polytope` that was generating an error if you tried to run `abstraction.discretize` with `closed_loop=False`.  The original hack was to set `closed_loop=True` on discretization, but this caused problems with the simulation.  With [polytope/PR #56](https://github.com/tulip-control/polytope/pull/56) in place, we can use `closed_loop=False` in both the discretization and in finding the control input.

* Added in the capability to pass the solver to `abstraction.discretize`.  This is used in some examples that we are preparing for the EECI course.

* The call `pc.is_inside(region, x0)` was causing errors with the number of arguments (2 vs 3).  Since we weren't passing a tolerance, I changed this to `x0 in region`.

* Some small docstring and error message fixes.

### DIFF
--- a/examples/continuous_with_simulation.py
+++ b/examples/continuous_with_simulation.py
@@ -90,10 +90,15 @@ print('\n Simulation starts \n')
 T = 100
 # let us pick an environment signal
 randParkSignal = [random.randint(0, 1) for b in range(1, T + 1)]
+
+# Set up parameters for get_input()
+disc_dynamics.disc_params['conservative'] = True
+disc_dynamics.disc_params['closed_loop'] = False
+
 # initialization:
 #     pick initial continuous state consistent with
 #     initial controller state (discrete)
-u, v, edge_data = ctrl.edges('Sinit', data=True)[1]
+u, v, edge_data = list(ctrl.edges('Sinit', data=True))[1]
 s0_part = edge_data['loc']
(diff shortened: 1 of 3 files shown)


murrayrm:
@johnyf pointed out in an e-mail that there is a change in [this commit](https://github.com/tulip-control/tulip-control/pull/230/commits/6b31f10185a832303aac53f6319c031e984b2baf#diff-f955a6e1ee45d31f8bf83b5056a5b65aR94-R96) that doesn't match the commit message.

The change is intentional, but it is indeed not in the right commit.  The change is related to make sure that `closed_loop=False` is used in both the discretization and when calling `get_input`.  Probably belongs is a separate commit message.  This could be cleaned up via a force push, but not sure it is worth the effort?  Let me know if you think I need to clean it up (but confirming that it was intentional, as mentioned above).

Alternative: if everything else looks OK, we could do a squash and merge and just get rid of the details commits.

johnyf:
I think that force pushing is not needed. The changes are ready for merging.

johnyf:
@murrayrm thank you for this pull request, it has been merged.

Issue #231, opened 2020-03-02
populate `__init__.py`


johnyf:
Add imports to `tulip/__init__.py` of modules and ensure that direct calls to functions work, for example `tulip.discretize` and `tulip.synthesize`.

murrayrm:
Interestingly, there is a line that imports the main modules in `__init__.py`, but it is commented out.  Does anyone remember why?

johnyf:
It appears to have been commented out to reduce runtime in the commit https://github.com/tulip-control/tulip-control/commit/7e05aeac30d3d24db2cbc00154c5b80ad2bc2743#diff-4505ab709ad2fcabf4c351b5c4084ea5R47, which changed some imports of `matplotlib` and `ipython` to be inline. `matplotlib` is now (ce54897c242689f45ad33650f157bf1805b35ed6) imported inline, except in the module `tulip.graphics`

https://github.com/tulip-control/tulip-control/blob/ce54897c242689f45ad33650f157bf1805b35ed6/tulip/graphics.py#L53-L58

slivingston:
I am reviewing PR #234, but I think that this issue is the best place for my question: is this change sufficiently motivated to justify the increase in time required to `import`?

As I understand it, populating the root `__init__.py` is motivated to make `import` statements shorter in practice. E.g., `from tulip import GRSpec` is shorter than `from tulip.spec import GRSpec`. Being shorter is better because it requires less time to write code and because it is easier to remember.

However, populating `__init__.py` has the disadvantage of increasing the time to import subpackages of TuLiP because all or most subpackages will always be imported, independently of whether they are used.

As a concrete example, with the changes of PR #234, the time required to execute `examples/gridworlds/genrand.py` is approximately 1.6 s. If I modify that PR to perform no imports in the root `__init__.py`, then `examples/gridworlds/genrand.py` executes within approximately 450 ms.

slivingston:
One counter-argument in favor of allowing the increased time to `import` is that for problems of nontrivial size, the difference of 1 or 2 seconds is not a significant proportion of total execution time.

However, small examples like `genrand.py` ([mentioned above](https://github.com/tulip-control/tulip-control/issues/231#issuecomment-594952997).) are nonetheless interesting for education and getting started with TuLiP.

slivingston:
To conclude, there are at least 3 options:
1. remove all internal imports from the root `__init__.py`,
[...]

PR #234, opened 2020-03-03
add imports to `__init__.py`, import `matplotlib` within functions, install parser tables

johnyf:
These changes:

- add more imports to `tulip/__init__.py` and the `__init__.py` of subpackages,
  to address issue #231. A few functions and classes are not included
  (even though their name does not start with an underscore),
  because they appeared to be internal to modules or are under development.

- move imports of `matplotlib` to within functions in the module `tulip.graphics` to
  reduce import time when importing `tulip`.

- load the parser tables from the generated module in `tulip.interfaces.ltl2ba`
  and generate the parser tables in `setup.py` when installing `tulip`.

- avoid a cyclic import in `tulip.spec.gr1_fragment` by importing within
  function `check` the subpackage `tulip.transys`.


slivingston:
Thanks for these changes. I was able to reproduce parser generator warnings and witness your fix (https://github.com/tulip-control/tulip-control/pull/234/commits/06a57810a86ace8208d405a93c789748701664cb), and I was able to reproduce a cyclic import on Python 2.7 that is avoided by the change in https://github.com/tulip-control/tulip-control/pull/234/commits/a41aca48b857674e39c42698aa1aae2f546f1e4b.

slivingston:
These changes are ready to merge, pending discussion in #231 regarding how many imports should be added to `__init__.py`.

murrayrm:
A slightly different way to do this, which is what we do in `python-control` is to use the `__all__` attribute within a module (file).  This lets you isolate the public functions to the file, which is perhaps a little bit cleaner?

johnyf:
I agree that it would be cleaner to use `__all__` in modules and `from subpackage import *`, `from module import *`. These are used in

- `python-control`: [`control/__init__.py`](https://github.com/python-control/python-control/blob/0.8.3/control/__init__.py), a [subpackage `__init__.py`](https://github.com/python-control/python-control/blob/0.8.3/control/matlab/__init__.py), and a module [using `__all__`](https://github.com/python-control/python-control/blob/0.8.3/control/bdalg.py#L61)

- `numpy`: [`numpy/__init__.py`](https://github.com/numpy/numpy/blob/v1.18.1/numpy/__init__.py#L143), a [subpackage `__init__.py`](https://github.com/numpy/numpy/blob/v1.18.1/numpy/core/__init__.py#L76-L120), and a module [using `_all__`](https://github.com/numpy/numpy/blob/v1.18.1/numpy/core/numeric.py#L52-L68)

- `networkx`: [`networkx/__init__.py`](https://github.com/networkx/networkx/blob/networkx-2.4/networkx/__init__.py#L95-L127), a [subpackage `__init__.py`](https://github.com/networkx/networkx/blob/networkx-2.4/networkx/generators/__init__.py), and a module [using `__all__`](https://github.com/networkx/networkx/blob/networkx-2.4/networkx/generators/atlas.py#L20)

`scipy` uses [explicit import for subpackages](https://github.com/scipy/scipy/blob/v1.2.3/scipy/__init__.py#L10-L16) and [`__all__`](https://github.com/scipy/scipy/blob/v1.2.3/scipy/fftpack/__init__.py#L88-L97) and [`import *`](https://github.com/scipy/scipy/blob/v1.2.3/scipy/fftpack/__init__.py#L99-L101) within subpackages, and also [`__all__ = [s for s in dir() if not s.startswith('_')]`](https://github.com/scipy/scipy/blob/v1.2.3/scipy/integrate/__init__.py#L97).
The pattern `__all__.extend(subpackage.__all__)` is [used in `numpy`](https://github.com/numpy/numpy/blob/0c32bb4140215f99b935eae9001eabb10e3c47dc/numpy/__init__.py#L169).

The approach in e8f65f7cd0a1758e587fb6f3748465254280501a of importing each function and class would require editing all the `__init__.py` files following any change in function or class name.

I will rewrite the `__init__.py` changes to use `__all__` in modules and `from subpackage import *`, `from module import *`.

slivingston:
To be clear, the question of using `__all__` or not is independent of my comment about performance.

PR #235, opened 2020-03-15
add function `simulate2d`

johnyf:
These changes:

- add the function `tulip.abstract.plot.simulate2d`, which is based on the
  code of the example `examples/continuous_with_simulation.py`.

- add the function `tulip.abstract.plot.pick_point_in_polytope`, which is
  based on the code of the example `examples/continuous_with_simulation.py`.

- add an example script `examples/continuous_forall_init.py`, which is
  based on the example `examples/continuous_with_simulation.py`,
  with initial condition quantification `\A \A`.

- conjoin system initial conditions to environment initial conditions for
  representing transition graphs in logic when using `\A \A` as quantification
  of initial conditions, because in this case the system initial conditions
  need to be empty or true.


johnyf:
Relevant to #146.

slivingston:
Excellent! Eventually we might want to return `x`, `y` from `simulate2d()`. (I am happy to wait until there is a motivating application to do this.)

PR #236, opened 2020-03-18
Minimum violation planning

tichakornw:
This PR adds the minimum-violation planning functionality to TuLiP. In particular, the following modules are introduced, with unit tests (tests/mvp_test.py, which includes an example from the EECI 2020 course).

* tulip.transys.cost: Cost Module for state or transition cost/weight. It provides VectorCost class for defining a vector cost where addition is performed element-wise and comparison is performed lexicographically. Additionally, ValidTransitionCost class is defined to allow cost with addition operation, 0 as its identity, and whose value is non-negative.

* tulip.transys.WeightedKripkeStructure and tulip.transys.WeightedFiniteAutomaton to define the weighted version (i.e., allowing "cost" attribute for edges) of KripkeStructure and FiniteAutomaton.

* tulip.transys.compositions module that include synchronous_parallel for constructing synchronous paralel composition of KripeStructure objects. Note that this is quite similar to the tensor_product function defined in tulip/transys/algorithms with the following difference
  - synchronous_parallel returns a weighted Kripke structure where cost on each edge is the sum of the cost of the corresponding transition on each model. tensor_product returns a labeled graph (so the information on the type is lost) and the cost on each edge will be the tuple of the cost of the corresponding transition on each model.
  - synchronous_parallel allows operation for each attribute to be specified (e.g., sum for cost and multiply for probability)

* tulip.spec.prioritized_safety module includes PrioritizedSpecification class for defining prioritized safety specification.

* tulip.transys.graph_algorithms for keeping algorithms, esp search algorithm, on graphs. Currently, it includes dijkstra_single_source_multiple_targets and dijkstra_multiple_sources_multiple_targets. The main difference from networkx search algorithm is that once a target inside targets set is found, the loop terminates.

* tulip.mvp is the module for solving minimum violation planning.

### DIFF
--- a/tests/mvp_test.py
+++ b/tests/mvp_test.py
@@ -1,0 +1,321 @@
+# This test includes unit tests for each element needed to formulate and solve
+# the minimum violation planning problem.
+# It also includes an integration test of minimum violation planning
+# using the example from the EECI 2020 computer lab.
+#
+# * vectorcost_test(): a unit test for tulip.transys.cost module
+# * wks_test(): a unit test for tulip.transys.WeightedKripkeStructure module
+# * wfa_test(): a unit test for tulip.transys.automata.WeightedFiniteStateAutomaton module
+# * composition_test(): a unit test for tulip.transys.compositions module
+# * prioritized_safety_test(): a unit test for tulip.spec.prioritized_safety module
+# * graph_algorithm_test(): a unit test for tulip.transys.graph_algorithms module
+# * mvp_test(): an integration test for minimum-violation planning based
+#   on the EECI 2020 computer lab.
+
+import copy
+from tulip.transys.cost import VectorCost
+from tulip.transys import WeightedKripkeStructure as WKS
+from tulip.transys.automata import WeightedFiniteStateAutomaton as WFA
+from tulip.transys.compositions import ks_synchronous_parallel
+from tulip.spec.prioritized_safety import PrioritizedSpecification
+from tulip.transys.mathset import PowerSet
+from tulip.transys.graph_algorithms import (
+    dijkstra_single_source_multiple_targets,
+    dijkstra_multiple_sources_multiple_targets,
+)
+from tulip.mvp import solve as solve_mvp
+
+
+def vectorcost_test():
+    num_item = 10
+    a = VectorCost([2 * i for i in range(num_item)])
+    assert len(a) == num_item
+    for i in range(num_item):
+        assert a[i] == 2 * i
+
+    b = a + 1
+    c = 1 + a
+    assert b >= a
+    assert b > a
+    assert a < b
+    assert a <= b
+    assert b == c
+    assert a != c
+    assert len(b) == num_item
+    assert len(c) == num_item
+
+    i = 0
+    for b_item in b:
+        assert b_item == 2 * i + 1
+        i += 1
+
+    d = VectorCost([2 * i if i != 2 else 2 * i + 1 for i in range(num_item)])
+
+    assert d > a
+
+
+def wks_test():
+    states = {"c8", "c4", "c9"}
+    transitions = {
+        ("c8", "c8", 1),
+        ("c8", "c4", 0),
+        ("c4", "c4", 1),
+        ("c4", "c9", 0),
+        ("c9", "c9", 0),
+    }
+    init = "c8"
+
+    ts = _construct_wks(states, transitions, init)
+
+    assert len(ts.states) == len(states)
+    for state in ts.states:
+        ts.states[state]["ap"] == {state}
+
+    for transition in transitions:
+        ts_transitions = ts.transitions.find(transition[0], transition[1])
+        assert len(ts_transitions) == 1
+        for ts_transition in ts_transitions:
+            assert ts_transition[2]["cost"] == transition[2]
+
+
+def wfa_test():
+    propositions = ["h4", "a4"]
+    fa = _construct_wfa(propositions, [[]])
+    assert len(fa.states) == 1
+    assert len(fa.transitions.find()) == 2 ** len(propositions)
+
+
+def composition_test():
+    # Test deterministic transitions
+    states1 = {"c1", "c2"}
+    transitions1 = {
+        ("c1", "c2", 1),
+        ("c2", "c1", 0),
+    }
+    init1 = "c1"
+
+    states2 = {"c1", "c2", "c3"}
+    transitions2 = {
+        ("c1", "c2"),
+        ("c2", "c3"),
+        ("c3", "c3"),
+    }
+    init2 = "c1"
+
+    states3 = {"c4", "c5"}
+    transitions3 = {
+        ("c4", "c5", 1),
+        ("c5", "c4", 0),
+    }
+    init3 = "c4"
+
+    ts1 = _construct_wks(states1, transitions1, init1)
+    ts2 = _construct_wks(states2, transitions2, init2)
+    ts3 = _construct_wks(states3, transitions3, init3)
+
+    ts = ks_synchronous_parallel([ts1, ts2, ts3])
+    assert isinstance(ts, WKS)
+
+    assert len(ts.states) == len(states1) * len(states2) * len(states3)
+    assert len(ts.states.initial) == 1
+    init = list(ts.states.initial)[0]
+    init_expected = (init1, init2, init3)
+
+    visited = {init}
+    visited_expected = {init_expected}
+    Q = [init]
+    Q_expected = [init_expected]
+
+    while len(Q) > 0:
+        v = Q.pop(0)
+        v_expected = Q_expected.pop(0)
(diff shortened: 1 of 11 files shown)


tichakorn832:
Travis is finally happy. @murrayrm , @slivingston Could you help take a look?

slivingston:
@tichakorn832 yes, I will review it today.

tichakorn832:
Thanks @slivingston. The part I'm most worried about is the code structure in terms of where each module should go. For example, the newly introducted `compositions` and `graph_algorithms` modules under `tulip/transys` might contain some duplication of the existing `algorithms` module under the same folder. (In my opinion, at some point "algorithms" will grow and we might want to separate them into different types of algorithms anyway.) Please let me know if you feel like different structures make more sense.

tests/mvp_test.py
```diff
@@ -0,0 +1,307 @@
+import copy
```
murrayrm: It would be nice to have some comments at the top of this file saying a bit about what it tests.

tulip/mvp.py
```diff
@@ -0,0 +1,196 @@
+# Copyright (c) 2012-2015 by California Institute of Technology
```
murrayrm: Year should be 2020.  Do you want to copyright as Caltech or UT Austin?

tulip/spec/prioritized_safety.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) 2013-2015 by California Institute of Technology
+# and 2014 The Regents of the University of Michigan
```
murrayrm: Remove this line.

tulip/spec/prioritized_safety.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) 2013-2015 by California Institute of Technology
```
murrayrm: Update?

tulip/spec/prioritized_safety.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def __iter__(self):
+        self._iter_level = 0
+        self._iter_index = 0
+        return self
+
+    def __next__(self):
+        while self._iter_level < len(self._Psi) and self._iter_index >= len(
+            self._Psi[self._iter_level]
+        ):
```
murrayrm: Is this PEP8 compliant?  Suggest running `pycodestyle`

tulip/transys/compositions.py
```diff
@@ -0,0 +1,150 @@
+# Copyright (c) 2013-2015 by California Institute of Technology
```
murrayrm: Update

tulip/transys/cost.py
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2013-2015 by California Institute of Technology
```
murrayrm: Update

tulip/transys/graph_algorithms.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) 2012-2015 by California Institute of Technology
```
murrayrm: Update

tulip/transys/graph_algorithms.py
```diff
@@ -0,0 +1,106 @@
...
+            if not current_cost or new_cost < current_cost:
+                dist[v] = new_cost
+                heappush(Q, (new_cost, v, path_to_u))
+
+    return (float("inf"), ())
+
+
+def dijkstra_multiple_sources_multiple_targets(
+    graph, source_set, target_set, cost_key="cost"
+):
```
murrayrm: Non-PEP8?

tulip/spec/prioritized_safety.py
tichakorn832: It was generated from `black` and validated with flake8.

tulip/transys/graph_algorithms.py
slivingston: right, not PEP8; cf. https://www.python.org/dev/peps/pep-0008/#indentation

tulip/spec/prioritized_safety.py
slivingston: As far as I know, it is PEP8 compliant, but some readers might find it difficult to read compared with the style of folding the line at the `and`, e.g.,
```python
while (self._iter_level < len(self._Psi)
       and self._iter_index >= len(self._Psi[self._iter_level])):
```

tulip/transys/graph_algorithms.py
tichakorn832: Neither pycodestyle or flake8 complains about this though. Are there other tools we can use to check? (This is also automatically formatted using black. Are there other tools you're using to automatically format?)

tests/mvp_test.py
tichakorn832: Done. The example from EECI is actually in this file (mvp_test() function). I think it's better to put it as part of tests instead of examples so that we keep checking that it always works.

tulip/mvp.py
tichakorn832: Fixed

tulip/spec/prioritized_safety.py
tichakorn832: Fixed

tulip/spec/prioritized_safety.py
tichakorn832: Fixed

tulip/transys/compositions.py
tichakorn832: Done

tulip/transys/cost.py
tichakorn832: Done

tulip/transys/graph_algorithms.py
tichakorn832: Done

tulip/transys/graph_algorithms.py
tichakorn832: I try using autopep8. It won't catch this. However, if it detects that the line is too long (I tried putting all these 3 lines into one line), then it would reformat as suggested in https://www.python.org/dev/peps/pep-0008/#indentation.

tichakorn832:
Hmm...after fixing the copyright statement, travis failed as a bunch of these are printed out in abstract_test.transition_directions_test. Any idea how this could happen? As far as I'm aware, this PR shouldn't touch anything that affects this test. In fact, abstract_test only imports `abstract`, `hybrid` and `polytope`. This PR doesn't touch any of these files.

```
8 rows, 3 columns, 16 non-zeros
*     0: obj =   0.000000000e+00 inf =   0.000e+00 (1)
*     3: obj =  -1.000000000e+00 inf =   0.000e+00 (0)
OPTIMAL LP SOLUTION FOUND
GLPK Simplex Optimizer, v4.57
4 rows, 3 columns, 8 non-zeros
*     0: obj =   0.000000000e+00 inf =   0.000e+00 (1)
*     3: obj =  -1.000000000e+00 inf =   0.000e+00 (0)
OPTIMAL LP SOLUTION FOUND
```

tulip/transys/graph_algorithms.py
slivingston: @tichakornw I do not know what is going on with style validation. For formatting, I use built-in Emacs `python-mode` and adjust manually sometimes. In particular, I am not using [black](https://github.com/psf/black).

I think your style here is OK, whether or not consistent with PEP 8.

slivingston:
[re:](https://github.com/tulip-control/tulip-control/pull/236#issuecomment-610292042)
> after fixing the copyright statement, travis failed as a bunch of these are printed out in abstract_test.transition_directions_test. Any idea how this could happen?

My hypothesis is that one of the dependencies changed in a way that breaks tests here, independently of your work on this pull request. I am investigating now.

johnyf:
The https://github.com/tulip-control/tulip-control/pull/242#issuecomment-648038944 may be relevant to the failing tests.

slivingston:
@johnyf perfect, thanks.

slivingston:
I am going to merge https://github.com/tulip-control/tulip-control/pull/242 first, then rebase this PR in order to leverage work from https://github.com/tulip-control/tulip-control/pull/242 regarding how `cvxopt` is installed.

tichakornw:
Thanks @slivingston and @johnyf.

slivingston:
@tichakornw thanks. Regarding the [question of where each module should go](https://github.com/tulip-control/tulip-control/pull/236#issuecomment-602436116), I think the layout you have here is good. We can iterate more later on the organization of tulip.transys.

Regarding [adding an example](https://github.com/tulip-control/tulip-control/pull/236#pullrequestreview-381722627), I think that the tests are sufficient for this pull request, and I do not want to cause more delay. Also, the [function mvp_test() contains a complete example](https://github.com/tulip-control/tulip-control/pull/236/commits/4daa5f97b89f5bf880408ddb241328dcd38dc979#diff-d616606878da9f79124f842900b24827R211).

Issue #169, opened 2016-11-13
consider using `nose2` instead of `nose`


johnyf:
`nose` is unmaintained, and its earlier maintainer recommends moving on to use `nose2` (see https://github.com/nose-devs/nose/pull/1004#issuecomment-260155457).

slivingston:
The homepage of `nose` lists several alternatives:
> New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

quoting from http://nose.readthedocs.io/en/latest/
where links are:
* nose2: https://github.com/nose-devs/nose2
* py.test: http://pytest.org/

johnyf:
`pytest` seems worth considering, it appears to be much more actively developed, and more widely used.

slivingston:
concerning differences between `nose` and `nose2`: https://nose2.readthedocs.io/en/latest/differences.html

slivingston:
In the tulip-control project, this issue also applies to

- [ ] [polytope](https://github.com/tulip-control/polytope)
- [ ] [dd](https://github.com/tulip-control/dd)

slivingston:
I recommend pytest and will move forward with it. This might be useful for legacy code using `nose`: https://docs.pytest.org/en/latest/nose.html

johnyf:
`nose` is not compatible with Python 3.10 (due to imports from the module [`collections`](https://docs.python.org/3.9/library/collections.html#module-collections) of classes [that have moved](https://docs.python.org/3.10/whatsnew/3.10.html#removed) to the module [`collections.abc`](https://docs.python.org/3.10/library/collections.abc.html#collections-abstract-base-classes)). I recently (in a local repository) transitioned `dd` from `nose` to `pytest`, which informed me about its usage. I would be happy to transition `tulip` to `pytest`. I also transitioned to `pytest` the packages `polytope` (on branch `py310`) and `omega` (in a local repository).

murrayrm:
Using pytest seems like a good path forward.

johnyf:
I will start the transition to using `pytest`.

johnyf:
The transition to `pytest` has been completed on branch `use_pytest`, not yet merged into the mainline branch. I also transitioned CI testing to GitHub Actions, as described in #247, because Travis CI is no more free.

So the branch `use_pytest` is I think nearly ready to be rebased and merged. I intend to wait a week or more, before merging this branch, in order to ensure that I have not forgotten anything.


## Some remarks

[...]

PR #237, opened 2020-03-22
Stormpy interface

tichakornw:
This PR builds on #236. (It makes some modification to synchronous parallel composition.)

It adds interface to stormpy by introducing the following classes/modules.

* tulip.transys.MarkovChain and tulip.transys.MarkovDecisionProcess to define a Markov chain and a Markov decision process. MarkovChain derives from KripkeStructure and adds "probability" to transition attribute. MarkovDecisionProcess derives from MarkovChain and adds "action" to transition attribute.

* Update ks_synchronous_parallel to synchronous_parallel and takes care of composition of Markov models.

* Add mathfunc module that includes FunctionOnLabeledState class. This class can be used, .e.g, to define a policy. Currently, it is used to store a policy generated by stormpy.

* tulip.interfaces.stormpy is the interface to stormpy. It contains functionality such as parsing tulip transys structure to prism file and vice versa.


tichakornw:
@slivingston @murrayrm 

I just rebased this PR with the current master. It's ready for review.

slivingston:
@tichakornw I plan to review it today

tests/stormpy_interface_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+# Set up paths to all the models
+model_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "models")
+light_path = os.path.join(model_path, "light.pm")
+ma_path = os.path.join(model_path, "ma.nm")
+mh_path = os.path.join(model_path, "mh.pm")
+out_model_path = os.path.join(model_path, "tmpout.nm")
+
+skip_test = False
+try:
```
slivingston, line 26: It is best for tests collected here to fail if `stormpy` is not installed, rather than automatically skip if `ImportError` is raised.

Because we use `./run_tests.py` to drive collection and performance of tests, the act of testing involves a decision about which tests to run. As such, we assume the user wants to perform a test if it is included by `./run_tests.py`, so failure should occur if dependencies are missing.

To be clear, the pattern here is a valid for organizing tests, but it is not consistent with the current pattern of Tulip testing.

slivingston: For example, like [tests/lily_test.py](https://github.com/tulip-control/tulip-control/blob/9b9a743d1486ded5a91eba2fbbb4536559b85f5f/tests/lily_test.py) and [tests/gr1pyint_test.py](https://github.com/tulip-control/tulip-control/blob/9b9a743d1486ded5a91eba2fbbb4536559b85f5f/tests/gr1pyint_test.py)

slivingston:
I want to add stormpy to CI builds. Because the installation involves some non-Python tools like `carl` and `Storm`, we should create a script that is known to work at least on Debian-like GNU/Linux hosts. E.g., [extern/get-lily.sh](https://github.com/tulip-control/tulip-control/blob/9b9a743d1486ded5a91eba2fbbb4536559b85f5f/extern/get-lily.sh).

I can make such a script during my review, following instructions at https://moves-rwth.github.io/stormpy/installation.html and https://moves-rwth.github.io/pycarl/installation.html

tests/stormpy_interface_test.py
tichakornw: I see. Thanks. I'll fix this.

tichakornw: Fixed in 6f868e1.

I also added new tests in this PR (except stormpy_interface_test) to run_test.py in 3291787.

tichakornw:
Yeah, installing all the dependencies for stormpy is not so simple. It will take a few hours. At the time I installed mine, which was around late Feb, I couldn't get the latest stormpy to work with the stable branch of storm, so I had to revert stormpy to commit 7ae4d0806edde02093d4f90ee25d381b344180ff. I wrote an instruction at the bottom of this page

https://www.cds.caltech.edu/~murray/wiki/index.php?title=EECI-IGSC_2020

regarding how to install carl, pycarl, z3, storm, and stormpy.

tests/stormpy_interface_test.py
tichakornw: It seems like even though I don't add stormpy_interface_test to run_tests, it's still run in travis. Do you know how to disable it?

slivingston:
Thanks for sharing your instructions. I will explore the possibility of [caching](https://docs.travis-ci.com/user/caching) the resulting installation on Travis CI.

tests/stormpy_interface_test.py
slivingston: Actually I want it to run on Travis. After I find a way to satisfy the `stormpy` dependency in Travis CI builds, then the tests will start to pass.

tichakornw:
Thanks @slivingston .

slivingston:
I almost have a script to build and install stormpy. My goal is to finish it and complete review of this PR tomorrow.

tichakornw:
Sounds good. Thanks @slivingston .

slivingston:
quick update: I became stuck at installing pycarl, but now it is working, so I will be able to finish tonight or tomorrow.

slivingston:
update: I now have a script that builds `stormpy`, Storm, etc. on Fedora and Ubuntu 18.04. I observed what you warned: building requires more than 1 hour. today I will try either the Travis CI cache or saving the Storm and CArL files somewhere and downloading for each CI job.

tichakornw:
Thank you @slivingston and sorry for the trouble.

Does Travis allow using a docker image? If so, we can create a docker image with these libraries installed.

slivingston:
@tichakornw yes, [Travis supports Docker](https://docs.travis-ci.com/user/docker/), and I have experience using it there, but it might not be possible to run Tulip tests outside a Docker container with dependencies inside a Docker container. 1 idea: for stormpy-related tests, Tulip is copied into the container, so Tulip tests are executed entirely inside there. I will try this if copying bin and lib files does not work.

No worries. This is not trouble. It is a journey and an opportunity!

slivingston:
I have a script that pulls built files of CArL and Storm from <https://sourceforge.net/projects/cudd-mirror/files/> and builds and installs pycarl and stormpy from these. In my local trials on Ubuntu 18.04 it works reliably, so now I am trying it on Travis CI with the temporary branch `stormpy-travis-ci`.

slivingston:
this is working on my [temporary branch](https://travis-ci.org/github/tulip-control/tulip-control/builds/741509890), so I am going to apply the changes onto the branch of this PR, and finally finish my review.

note that [stormpy](https://moves-rwth.github.io/stormpy/) requires Python 3, whereas Tulip still supports Python 2.7. @tichakornw do you have a recommended way to proceed?

my proposal is
1. skip the tests if Python major version number is 2
2. check the Python major version number when users import `interfaces.stormpy` and, if 2, print warning

the disadvantage of item 2 is it incurs a small, unnecessary cost for users of Python 3, which seems unfair because Python 3 is what everyone should be using anyway. therefore, we could also skip item 2 and let it crash for users of Python 2 who try the stormpy interface

tichakornw:
Thanks, @slivingston. I thought that with the removal of python 2 support (which should be coming soon), 1 might be sufficient. What do you think?

tests/mdp_test.py
```diff
@@ -0,0 +1,92 @@
+# This test includes unit tests for MarkovChain and MarkovDecisionProcess of
```
slivingston, line 1: It is a best practice (recommended by [PEP 257](https://www.python.org/dev/peps/pep-0257/)) to have a docstring at the start of a file instead of a comment block because this exposes the documentation to tools that read the `__doc__` attribute of modules. E.g., we do this in [transys_automata_test.py](https://github.com/tulip-control/tulip-control/blob/9b9a743d1486ded5a91eba2fbbb4536559b85f5f/tests/transys_automata_test.py).

However, I think this can be merged as-is and changed later because this is a file of tests, rather than a part of the `tulip` package.

tests/stormpy_interface_test.py
```diff
@@ -0,0 +1,184 @@
...
+if not skip_test:
+    from tulip.transys.compositions import synchronous_parallel, apply_policy
+    from tulip.interfaces import stormpy as stormpy_int
+
+# Set up paths to all the models
+model_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "models")
+light_path = os.path.join(model_path, "light.pm")
+ma_path = os.path.join(model_path, "ma.nm")
+mh_path = os.path.join(model_path, "mh.pm")
+out_model_path = os.path.join(model_path, "tmpout.nm")
```
slivingston, line 32: It is best to use [mkstemp](https://docs.python.org/3.7/library/tempfile.html#tempfile.mkstemp) to create a temporary file according to standard OS calls or to use a temporary directory fixture provided by the test runner. The latter will be [available in pytest](https://docs.pytest.org/en/stable/tmpdir.html) after we switch to it (task from issue #169).

However, there is code that follows the same pattern as here [in gr1cint_test.py](https://github.com/tulip-control/tulip-control/blob/9b9a743d1486ded5a91eba2fbbb4536559b85f5f/tests/gr1cint_test.py#L171). It has not caused trouble in practice yet, so I think we can merge the code here and possibly update to mkstemp() during work on #169

slivingston:
@tichakornw this interface with Stormpy is exciting in general because it introduces probabilistic objects into Tulip, and in particular because interacting with [Storm](https://www.stormchecker.org/) is well motivated.

re: https://github.com/tulip-control/tulip-control/pull/237#issuecomment-722053959 yes, I agree that item 1 ("skip the tests if Python major version number is 2") is sufficient.

slivingston:
Adding Stormpy on Travis increases CI job total durations by approximately 3 times; e.g., 4 minutes 52 s of the Python 2.7 job vs. 15 minutes 3 s of the Python 3.8 job at https://travis-ci.org/github/tulip-control/tulip-control/builds/741525664

We can make this faster by compiling Stormpy and pycarl and saving the files [as done for Storm and CArL](https://github.com/tulip-control/tulip-control/pull/237#issuecomment-721913076). However, I am not available to work on that yet, so I am going to go ahead and merge.

tests/mdp_test.py
tichakornw: Thanks @slivingston. Sorry I didn't get a chance to address this review. I'll revisit these comments after the semester is over.

tests/stormpy_interface_test.py
tichakornw: Very nice suggestion. I didn't know about this. I'll revisit this as well as the previous comment.

PR #242, opened 2020-06-17
Fix find_controller for target regions with several polytopes

krooken:
In `tulip.abstract.find_controller.get_input()`.

The end state when finding a control action between two discrete regions might consist of several polytopes. For some of the polytopes in the target region there might not be a control action that brings the system there. In that case the matrix constructed by get_input_helper will be singular and the LP solver cannot return a solution. This is not a problem unless all polytopes in the end region are unreachable, in which case it seems likely that there is something wrong with the abstraction routine.

For instance, in the following example (as set up in the commit fe6eb342b94465cb37135a4506a9a2a0c1f4ab92) a controller input exists to bring the system from (0.5, 0.0) in region 0 to a point in region 1 (as shown by the arrows). However, there is no such controller input that brings the system to the left polytope of region 1.
partition

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
   - ./run_tests.py --outofsource --fast hybrid
   # install cvxopt.glpk
   - export CVXOPT_BUILD_GLPK=1
-  - pip install cvxopt==1.1.8
+  - python -c "import setup; setup.install_cvxopt()"
   - python -c "import cvxopt.glpk"
   # install `dd.cudd`
   - pip uninstall --yes dd
--- a/tulip/abstract/find_controller.py
+++ b/tulip/abstract/find_controller.py
@@ -269,6 +269,7 @@ def get_input(
 
         # for each polytope in target region
         for P3 in P_end:
+            cost = np.inf
             if mid_weight > 0:
                 rc, xc = pc.cheby_ball(P3)
                 R[
@@ -279,10 +280,27 @@ def get_input(
                 ] += mid_weight * np.eye(n)
 
                 r[idx, 0] += -mid_weight * xc
-                u, cost = get_input_helper(
-                    x0, ssys, P1, P3, N, R, r, Q, ord,
-                    closed_loop=closed_loop, solver=solver
-                )
+                try:
+                    u, cost = get_input_helper(
+                        x0, ssys, P1, P3, N, R, r, Q, ord,
+                        closed_loop=closed_loop, solver=solver
+                    )
+                except _InputHelperLPException as ex:
+                    # The end state might consist of several polytopes.
+                    # For some of them there might not be a control action that
+                    # brings the system there. In that case the matrix
+                    # constructed by get_input_helper will be singular and the
+                    # LP solver cannot return a solution.
+                    # This is not a problem unless all polytopes in the end
+                    # region are unreachable, in which case it seems likely that
+                    # there is something wrong with the abstraction routine.
+                    logger.info(repr(ex))
+                    logger.info((
+                        "Failed to find control action from continuous "
+                        "state {x0} in discrete state {start} "
+                        "to a target polytope in the discrete state {end}.\n"
+                        "Target polytope:\n{P3}").format(
+                            x0=x0, start=start, end=end, P3=P3))
                 r[idx, 0] += mid_weight * xc
 
             if cost < low_cost:
@@ -427,7 +445,7 @@ def get_input_helper(
                 "solver specified but only 'None' allowed for ord = 2")
         sol = solvers.qp(P, q, G, h)
         if sol['status'] != "optimal":
-            raise Exception(
+            raise _InputHelperQPException(
                 "getInputHelper: "
                 "QP solver finished with status " +
                 str(sol['status']))
@@ -452,7 +470,7 @@ def get_input_helper(
         h_LP = np.vstack((np.zeros((2 * N * (n + m), 1)), M))
     sol = pc.polytope.lpsolve(c_LP.flatten(), G_LP, h_LP, solver=solver)
     if sol['status'] != 0:
-        raise Exception(
+        raise _InputHelperLPException(
             "getInputHelper: "
             "LP solver finished with error code " +
             str(sol['status']))
@@ -460,6 +478,14 @@ def get_input_helper(
     u = var[-N * m:]
     cost = sol['fun']
     return u.reshape(N, m), cost
+
+
+class _InputHelperLPException(Exception):
+    pass
+
+
+class _InputHelperQPException(Exception):
+    pass
 
 
 def is_seq_inside(x0, u_seq, ssys, P0, P1):
--- a/setup.py
+++ b/setup.py
@@ -125,5 +125,31 @@ def run_setup():
               '!' * 65)
 
 
+def install_cvxopt():
+    """Install a version of cvxopt that is compatible with polytope
+    requirements"""
+    import polytope
+    import subprocess
+    import sys
+    ver = polytope.__version__
+    # Download all files for the current version of polytope
+    subprocess.check_call([
+        sys.executable, "-m",
+        "pip", "download",
+        "-d", "temp",
+        "--no-deps",
+        "polytope=={ver}".format(ver=ver)])
+    # Extract the tar archive
+    subprocess.check_call([
+        "tar", "xzf",
+        "temp/polytope-{ver}.tar.gz".format(ver=ver),
+        "-C", "temp"])
+    # Install cvxopt according to requirements file provided by polytope
+    subprocess.check_call([
+        sys.executable, "-m",
+        "pip", "install",
+        "-r", "temp/polytope-{ver}/requirements/extras.txt".format(ver=ver)])
+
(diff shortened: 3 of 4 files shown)


krooken:
All Travis nodes fails because the log output is too long:
```
The job exceeded the maximum log length, and has been terminated.
```
I do not know if this is expected or not, but these failures do not seem to be related to the PR.

murrayrm:
Not sure why this changed from previous tests, but it looks like the Travis CI script is doing a bunch of different variants of installed packages to make sure everything works.  We may need to split those into a test matrix so that each test case is a more manageable size.

krooken:
I think the problem only has to do with GLPK. It spews out 99701 lines of results from (it seems) 17000 different function calls.

I think I know what has gone wrong. The Travis jobs started to fail in April. I have not checked all of them, but the random subset of jobs that I have checked all fails because of an excessive amount of printouts from GLPK. One thing that happened in late March was that `polytope` was bumped from version `0.2.1` to `0.2.2`. In that range of commits is one commit that is a workaround for a bug in `cvxopt-1.2.0`: tulip-control/polytope@2672e75954dc358d11153f5ef36417feaf84c784. The bug has to do with suppression of GLPK output. TuLiP, however, is using `cvxopt-1.1.8` for which the output suppression looks differently: cvxopt/cvxopt#63.

The logs do not tell me which version of polytope is used in the jobs, but it seems likely that pip pulls the latest version of polytope from PyPI.

krooken:
Travis is now green and happy, so the configuration update that makes Travis install `cvxopt-1.2.0` seems to resolve the issue. Is it ok to have that slightly unrelated commit in this PR, or should I make a new one?

tulip/abstract/find_controller.py
```diff
@@ -279,10 +279,30 @@ def get_input(
...
+                    # there is something wrong with the abstraction routine.
+                    logger.info(repr(ex))
+                    logger.info(
+                        "Failed to find control action from continuous " + \
+                        "state " + str(x0) + " in discrete state " + \
+                        str(start) + " to a target polytope in the " + \
+                        "discrete state " + str(end) + ".\n" + \
+                        "Target polytope:\n" + \
+                        str(P3))
+                    u = 0.0
```
johnyf, line 304: From the assignment `low_u = np.zeros([N, m])` [above](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR268) the dimension of `u` seems to be `[N, m]`.

`low_cost` is the minimal value of cost over the polytopes that have been iterated over, because initially `low_cost = np.inf` ([above](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR267) before the iteration) and `low_cost` changes to only smaller values [below](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR308) (if `cost < low_cost`).

On the [next line](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR305) `cost = np.inf`, so `cost < low_cost` is `False` [below](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR308), and this value of `u` is unused. So `u` could also be left undefined here.

tulip/abstract/find_controller.py
```diff
@@ -279,10 +279,30 @@ def get_input(
...
-                u, cost = get_input_helper(
-                    x0, ssys, P1, P3, N, R, r, Q, ord,
-                    closed_loop=closed_loop, solver=solver
-                )
+                try:
+                    u, cost = get_input_helper(
+                        x0, ssys, P1, P3, N, R, r, Q, ord,
+                        closed_loop=closed_loop, solver=solver
+                    )
+                except Exception as ex:
```
johnyf, line 287: Catching `Exception` would include all exceptions, also those due to reasons other than unreachability of the target polytope (for example [here](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR445-R447)).

The function `get_input_helper` raises `Exception` when the QP solution status is not `'optimal'` ([here](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR450-R453)) ([documentation of the `cvxopt` function `qp`](https://cvxopt.org/userguide/coneprog.html#cvxopt.solvers.qp)) and when the LP solution status is not 0 ([here](https://github.com/tulip-control/tulip-control/pull/242/files#diff-db73a44c4631fa5fc7c04bff0f7d729aR475-R478)) ([cases of solution status when using the `cvxopt` function `lp`](https://github.com/tulip-control/polytope/blob/ccb080dbae7ffbb46e2ebe016665d9d0895fc1cd/polytope/solvers.py#L111-L118) and [documentation of the function `scipy.optimize.linprog`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html)).

A possibility could be to subclass [`Exception`](https://docs.python.org/3/library/exceptions.html#Exception), for example:

```python
class _GetInputHelperError(Exception):
    """Could not solve the given problem."""
    pass
```

and to `raise _GetInputHelperError` from within the function `get_input_helper` to signify errors in the solution status. This would enable catching `_GetInputHelperError` here, and raising other errors.

tulip/abstract/find_controller.py
```diff
@@ -279,10 +279,30 @@ def get_input(
...
+                    # The end state might consist of several polytopes.
+                    # For some of them there might not be a control action that
+                    # brings the system there. In that case the matrix
+                    # constructed by get_input_helper will be singular and the
+                    # LP solver cannot return a solution.
+                    # This is not a problem unless all polytopes in the end
+                    # region are unreachable, in which case it seems likely that
+                    # there is something wrong with the abstraction routine.
+                    logger.info(repr(ex))
+                    logger.info(
```
johnyf, line 297: An equivalent way to write this statement could be

```python
logger.info((
    'Failed to find control action from continuous '
    'state {x0} in discrete state {start} '
    'to a target polytope in the discrete state {end}.\n'
    'Target polytope:\n'
    '{P3}').format(
        x0=x0, start=start, end=end, P3=P3))
```

.travis.yml
```diff
@@ -39,7 +39,7 @@ script:
   - ./run_tests.py --outofsource --fast hybrid
   # install cvxopt.glpk
   - export CVXOPT_BUILD_GLPK=1
-  - pip install cvxopt==1.1.8
+  - pip install cvxopt==1.2.0
```
johnyf, line 42: [On Travis CI `polytope` installs the latest `cvxopt` version](https://github.com/tulip-control/polytope/blob/ccb080dbae7ffbb46e2ebe016665d9d0895fc1cd/.travis.yml#L43) and [in `requirements/extras.txt` `polytope` specifies `cvxopt==1.2.4`](https://github.com/tulip-control/polytope/blob/ccb080dbae7ffbb46e2ebe016665d9d0895fc1cd/requirements/extras.txt#L1). The latest `cvxopt` version is [1.2.5](https://pypi.org/project/cvxopt/#history). A number of possibilities would be to unpin `cvxopt` here, or pin `cvxopt` for `polytope` on Travis CI.

krooken: It seems most reasonable to let `polytope` decide what version of `cvxopt` that should be installed. After all, overriding `polytope`'s requirements led to this problem in the first place.

So, removing lines 41 and 42 entirely would make pip look at the requirements of `polytope`? Do we risk not getting GLPK in that case?

johnyf: `cvxopt` is not listed in the requirements of [`polytope`'s `setup.py`](https://github.com/tulip-control/polytope/blob/ccb080dbae7ffbb46e2ebe016665d9d0895fc1cd/setup.py) because it is not a default requirement. So when `pip` installs `polytope` it will not automatically install `cvxopt`, and lines 41 and 42 would be needed. If the latest version of `cvxopt` is installed here, then line 42 could become `pip install cvxopt`.

krooken: Right, I am learning of how `pip`, Travis and `polytope` are set up as we go. Using latest version or hard-coding it to the same version as `polytope` has right now would solve the problem, but it will open up for the same problem again in the future. We could pin the versions of both `polytope` and `cvxopt`, but then we put unnecessary restrictions on users of TuLiP who do not use `cvxopt`. The ideal solution would be to fetch the correct version of `extras.txt` and feed it to `pip`. In that way `polytope` defines which version of `cvxopt` to use, and we do not need to keep track of the correct version in TuLiP too.

krooken:
I've now addressed the review comments and Travis is again green and happy.

johnyf:
These changes look good to me.

murrayrm:
Thanks @krooken and @johnyf.  I'll leave this sitting for a day or two, just in case others have comments, but then go ahead and merge if there are no further suggestions.

slivingston:
@krooken thanks for these improvements. @johnyf @murrayrm thanks for reviewing.

Issue #241, opened 2020-06-12
confusing assertion error


necozay:
It looks like when a spec is trivially realizable by rendering the environment initial condition or environment safety false, the ``synthesize`` function, with default options, returns an assertion error: e.g., ```AssertionError: removed all NUMBER nodes!```. It looks like in this case the solver generates an automaton where all paths lead to a dead end and trimming this automaton returns an empty controller, hence the error. However this error message is confusing for new users. I  suggest updating the error to something like ```AssertionError: removed all NUMBER nodes! Please check env_init and env_safety to avoid trivial realizability.```  If you have any thoughts on this (and on how to make the error message more informative), please let me know. Otherwise, I will send in a pull request in a week.

Minimal example:
```
env_vars = {'e'}
env_init = set()                
env_init |= {'e'}
env_prog = set()
env_safe = {'e -> X(! e)', '!e -> s'}
sys_vars = {'s'}
sys_init = set() 
sys_prog = set() 
sys_safe = set() 
specs = spec.GRSpec(env_vars, sys_vars, env_init, sys_init,
                    env_safe, sys_safe, env_prog, sys_prog)
specs.qinit = '\A \E'  # Moore initial condition synthesized too
specs.moore = False
specs.plus_one = False
ctrl = synth.synthesize(specs, solver='gr1c', rm_deadends = 1)
```

murrayrm:
Definitely a good idea to add some more information to the error message. @tichakorn832 and I noticed that there are some pretty opaque messages that get returned.

slivingston:
done in https://github.com/tulip-control/tulip-control/pull/243

PR #243, opened 2020-06-21
improve error message in case of trivial realizability

necozay:
addresses the issue #241

### DIFF
--- a/tulip/transys/labeled_graphs.py
+++ b/tulip/transys/labeled_graphs.py
@@ -1117,7 +1117,11 @@ class LabeledDiGraph(nx.MultiDiGraph):
                  if d == 0}
             self.states.remove_from(s)
         m = len(self)
-        assert n == 0 or m > 0, 'removed all {n} nodes!'.format(n=n)
+        assert n == 0 or m > 0, ( 
+            'removed all {n} nodes!'.format(n=n) + '\n'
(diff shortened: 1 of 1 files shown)


tulip/transys/labeled_graphs.py
```diff
@@ -1117,7 +1117,11 @@ def remove_deadends(self):
                  if d == 0}
             self.states.remove_from(s)
         m = len(self)
-        assert n == 0 or m > 0, 'removed all {n} nodes!'.format(n=n)
+        assert n == 0 or m > 0, ( 
+            'removed all {n} nodes!'.format(n=n) + '\n'
+            ' Please check env_init and env_safety to avoid trivial'
+            ' realizability. Alternatively, you can set "rm_deadends = 0"'
```
slivingston, line 1123: In the documentation of `synthesize()`, we write that the type of `rm_deadends` is `bool`, whereas this message suggests using 0. At runtime this is not important because `rm_deadends` is used in a branch condition like `if rm_deadends`, so the consequence is the same whether 0 or `False`.

slivingston:
the failure of the CI jobs is not related to this PR. my 1 comment is trivial and can be handled later or skipped.

@necozay thanks! This PR clearly addresses the original issue #241

PR #245, opened 2020-10-02
FIX: Typo

krooken:
`solvers` is an imported module and not an argument to the function.


slivingston:
@krooken good catch! The change that introduced `if solvers` is from PR #230. We should eventually make a test with `get_input()` and `ord=2`.

slivingston:
Accepted via `git cherry-pick`: https://github.com/tulip-control/tulip-control/commit/9b9a743d1486ded5a91eba2fbbb4536559b85f5f

PR #254, opened 2023-12-06
DOC: recommend paper to cite

slivingston:
I just read a paper published this year that cites a paper about Tulip from 13 years ago. Presumably the authors did not know that there is a more recent paper. This pull request seeks to encourage users to cite it.

Note that the conference name ("IEEE Conference on Control Applications (CCA)") is as indicated in IEEExplore at <https://ieeexplore.ieee.org/document/7587949>. However, this is different than in the reference we have in our docs: https://github.com/tulip-control/tulip-control/blob/b983684c92cdd71bbed8c2765a7766b4868ce54f/doc/bib.txt#L38

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,22 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+url: "https://github.com/tulip-control/tulip-control"
+preferred-citation:
+  type: conference-paper
+  title: "Control design for hybrid systems with TuLiP: The Temporal Logic Planning toolbox"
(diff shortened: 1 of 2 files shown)


slivingston:
Guide to the Citation File Format: https://github.com/citation-file-format/citation-file-format/blob/e10d936e818ed7fdf18c0d3659989129ab8bd7b4/schema-guide.md

This is parsed by GitHub as described at <https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files>.

PR #256, opened 2024-02-20
refactoring, type hints, bugfixes

johnyf:
Changes including:

- refactoring of code
- updates to docstrings
- type hints and file `py.typed`
- bugfixes

### DIFF
--- a/tulip/abstract/prop2partition.py
+++ b/tulip/abstract/prop2partition.py
@@ -107,7 +107,6 @@ class PropPreservingPartition(
                 bool=True):
         if regions is None:
             regions = list()
-        # super().__init__(adj)
         if prop_regions is None:
             self.prop_regions = None
         else:


tulip/gridworld.py
```diff
@@ -326,7 +326,8 @@ def is_reachable(
         @return:
             True if there is a path, False otherwise.
         """
-        # Check input values and handle negative coordinates
+        # Check input values and
```
slivingston, line 353: Can you motivate such short lines? I find that readability is less with these.

slivingston: My only concern is having line widths that are different between comments and code. However, I am happy to keep these changes and test usefulness in practice and get feedback from others later (after merging).

tests/dumpsmach_test.py
```diff
@@ -64,8 +61,12 @@ def test_nx():
...
     print(dumpsmach.python_case(g, classname='Machine', start='Sinit'))
     exe_globals = dict()
-    exec(dumpsmach.python_case(g, classname='Machine', start='Sinit'), exe_globals)
-    m = exe_globals['Machine']()  # previous line creates the class `Machine`
+    exec(
+        dumpsmach.python_case(
+            g, classname='Machine', start='Sinit'),
+        exe_globals)
+    m = exe_globals['Machine']()
+        # previous line creates the class `Machine`
```
slivingston, line 69: Moving "previous line creates the class `Machine`" to the line after `exe_globals['Machine']()` is confusing because "previous line" refers to the line with `exec()`. I recommend to move it up a line, i.e.,
```suggestion
        exe_globals)
        # previous line creates the class `Machine`
    m = exe_globals['Machine']()
```

contrib/parity_games/solve_parity_game.py
```diff
@@ -1,110 +1,107 @@
...
+"""McNaughton's algorithm for solving parity games.
+
 
 Reference
 =========
+
 Robert McNaughton (1993). Infinite games played on finite graphs.
 Annals of Pure and Applied Logic 65(2): 149--184.
-doi:10.1016/0168-0072(93)90036-D
+<doi:10.1016/0168-0072(93)90036-D>
```
slivingston, line 9:
```suggestion
<https://doi.org/10.1016/0168-0072(93)90036-D>
```
or `[doi:10.1016/0168-0072(93)90036-D](https://doi.org/10.1016/0168-0072(93)90036-D)`. Anyway, `doi` is not a scheme recognized by Chrome nor Firefox. When I try `doi:10.1016/0168-0072(93)90036-D` in both browsers, they search for it, and the top result from DuckDuckGo has the title "Carcass feeding of captive cheetahs".

contrib/readdstar.py
```diff
@@ -141,23 +145,24 @@ def readdstar(getline):
...
-                break  # file.readline() returns empty string at EOF
+                break
+                    # file.readline() returns empty string at EOF
             parts = line.split()
             if len(parts) == 0:
-                continue  # Ignore blank lines
-
-            if not parts[0].endswith(":") and len(parts) == 3:  # id line
+                continue
+                    # Ignore blank lines
```
slivingston, line 160: Is there a style guide that recommends this pattern of putting comments on the next line with more indentation? I do not see how it is an improvement, but I am willing to try it.

tulip/interfaces/gr1c.py
```diff
@@ -29,75 +29,95 @@
...
 
-def check_gr1c():
-    """Return `True` if `gr1c >= require_version` found in PATH."""
+Tag = (
+    str |
+    ET.Element)
+_Cast = (
+    _abc.Callable |
+    None)
+_VarDom = (
```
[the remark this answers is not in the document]
slivingston: Thanks! That is good motivation.

tulip/transys/transys.py
```diff
@@ -709,90 +925,122 @@ def line_labeled_with(L, m=0):
...
+    any label in `ap_labels`.
 
-    For example if isinstance(ofts, FTS):
+    For example if `isinstance(ofts, FTS)`:
 
-      >>> from tulip.transys.transys import add_initial_states
-      >>> initial_labels = [{'home'}]
-      >>> add_initial_states(ofts, initial_labels)
+    ```python
+    import tulip.transys.transys as _trs
```
slivingston, line 994: From a user's perspective (i.e., the perspective of a user who reads this docstring), the name prefix `_` is not motivated, and it is easier to use `trs`:
```python
import tulip.transys.transys as trs
initial_labels = [{'home'}]
trs.add_initial_states(ofts, initial_labels)
```

tulip/transys/transys.py
```diff
@@ -928,53 +1229,77 @@ def _output_fts(ts, transitions, sol):
...
+
+    Create a bi/dual-simulation abstraction for
+    the finite transition system `ts`.
+
+    @param ts:
+        input finite transition system,
+        the one you want to get
+        its bi/dual-simulation abstraction.
+    @param simu_type:
+        string 'bi'/'dual', flag used to switch b.w.
```
slivingston, line 1262:
```suggestion
        string 'bi'/'dual', flag used to switch between
```
Not the fault of this PR, but the abbreviation "b.w." for "between" is not common, so I recommend that it is expanded.

tulip/abstract/prop2partition.py
```diff
@@ -30,83 +30,305 @@
...
+                sp.lil_array |
+                None=None,
+            prop_regions:
+                dict[..., Polytope] |
+                None=None,
+            check:
+                bool=True):
+        if regions is None:
+            regions = list()
+        # super().__init__(adj)
```
slivingston, line 110: This comment can be deleted because `super().__init__()` is called below and the original comment is from commit 3d3383beecce942a6499e3ee0f2d8e953e1ec3b2, not relevant anymore.

slivingston (approved):
Excellent work! Thanks for adding type annotations and for generally moving us toward better static analysis. Thanks also for many style improvements, deletion of Python 2 compatibility code, and bugfixes.

I made several comments that I hope you will consider before merging, but none of these are required. A small percentage of changes in your pull request are style changes for which I can imagine good arguments for/against. I am happy to just accept these and experience the style in practice.

Now that my review is done, I am happy with the decision to have Python version 3.10 as the minimum. When I began review, this was not clear to me, so for posterity, here is motivation:

Python 3.10 allows us to use (1) better type syntax and (2) structural pattern matching (new `match` and `case` statements). However, for comparison, the earliest Python version supported by the current release of NumPy (https://pypi.org/project/numpy/) is 3.9. For NumPy 1.24 (the minimum required by Tulip), Python version 3.8 is feasible. Kera, a popular project in machine learning research, has minimum Python version 3.9 (https://github.com/keras-team/keras/blob/9d58974a94964aba4deb5b16f1bec2cde3330135/setup.py#L51).

Nonetheless, [SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends dropping 3.9 support. For example, this guidance is [followed in the Python Control Systems Library](https://github.com/python-control/python-control/pull/961).

slivingston:
Thanks for handling the comments and finding several more small items to fix. While reviewing this, I was reminded that we need to increase test coverage. I will begin to write more tests after the next release.

I am going to make a release of `polytope`; then, the CI builds of `tulip` should be passing again.

PR #261, opened 2024-03-15
Create pyproject.toml; fix layout error (in some browsers)

### LINKED ISSUE #260: README is not used as description on PyPI
I just made a release <https://pypi.org/project/tulip/1.4.0/> that resulted in
> The author of this package has not provided a project description

where a copy of the Tulip README should appear. Unlike the old version of PyPI, we cannot manually edit this description text for a release (cf. <https://github.com/pypi/warehouse/issues/4816> and references therein). Therefore, the best solution is to fix it for the next release. I expect this to involve copying README.rst into `long_description` of `setuptools.setup`.

slivingston:
## Summary

This pull request creates a pyproject.toml file that is a direct translation of parts of setup.py, so the resulting behavior should be the same. I am working on making the installation process of TuLiP easy, and changes to pyproject.toml may be required soon. Nonetheless I recommend that we commit pyproject.toml as it is here to provide a baseline for improvements.

This pull request also includes a layout fix for the User's Guide where the sidebar collapse button is misaligned. You should be able to see it at <https://tulip-control.github.io/tulip-control/> with a Firefox browser. 

## What is pyproject.toml ?

The Python packaging ecosystem has evolved a lot over the years. Because the needs of the community do not seem to be amenable to a single universal tool, there is now a standard file for expressing project data and build system requirements that projects should use independently of their tool choices: [pyproject.toml](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/).

Because setuptools is so popular, there are straightforward guides for translating from setup.py as we used it in TuLiP: <https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#writing-pyproject-toml> and <https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html>.

## Related issues

close #260

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,59 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "tulip"
+description = "Temporal Logic Planning (TuLiP) Toolbox"
+authors = [{name = "Caltech Control and Dynamical Systems", email = "tulip@tulip-control.org"}]
+readme = "README.rst"
+license = {text = "BSD-3-Clause"}
+classifiers = [
+    "License :: OSI Approved :: BSD License",
+    "Development Status :: 2 - Pre-Alpha",
+    "Intended Audience :: Science/Research",
+    "Topic :: Scientific/Engineering",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3 :: Only",
+]
+dynamic = ["version"]
+requires-python = ">=3.10"
+dependencies = [
+    "graphviz >= 0.20",
+    "networkx >= 2.0",
+    "numpy >= 1.24",
+    "omega >= 0.3.1, < 0.4.0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -0,0 +1,59 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "tulip"
+description = "Temporal Logic Planning (TuLiP) Toolbox"
+authors = [{name = "Caltech Control and Dynamical Systems", email = "tulip@tulip-control.org"}]
```
slivingston, line 8: We may want to change this authorship to be something more generic now. For comparison, [NumPy has "NumPy Developers"](https://github.com/numpy/numpy/blob/dc33ac4e1e59e49841f03b02098e73ba422ba56d/pyproject.toml#L17), and the [Python Control Systems Library has "Python Control Developers"](https://github.com/python-control/python-control/blob/7a70be1b7782abe0df0c69fc5d960ceb328ffbc2/pyproject.toml#L12).

slivingston: This can be fixed in the next pull request about documentation and installation.

PR #266, opened 2025-12-05
MNT: use `packaging`, instead of `pkg_resources`

johnyf:
- In `tests/version_test.py` use the package [`packaging`](https://pypi.org/project/packaging/), instead of [`pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html) (of `setuptools`), because `pkg_resources` is deprecated.

- Replace usage of `pkg_resources` within `tulip` without adding dependencies.


johnyf:
I intend to merge.

PR #268, opened 2025-12-29
CI: updates, Python 3.14, BUG: specific token values, REF: `TypeAlias`

johnyf:
The changes include:

- CI: test using CPython 3.12, 3.13, 3.14
- CI: install dependencies to build `cvxopt` wheel for Python 3.14
- CI: update versions of GitHub Actions used
- CI: use Ubuntu 24.04 (was 22.04)
- REF: omit `typing.TypeAlias`, which was deprecated in Python 3.12
- REL: use string as value for `project.license` and omit relevant Trove classifier, because it raises a `setuptools` error once `project.license` becomes a string value
- MNT: for editable installs use `pip install -e .` (was `python setup.py develop`)
- BUG: the lexemes `~`, `=>`, `/=`, `=<`, `<=>` propagated beyond the lexer, but instead the values `!`, `->`, `!=`, `<=`, `<->` were expected by later stages. Changed the lexer to set the value in each case to `!`, `->`, `!=`, `<=`, or `<->`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,7 +72,7 @@ jobs:
         if: github.ref == 'refs/heads/main'
         strategy:
             matrix:
-                python-version: ['3.10']
+                python-version: ['3.14']
         steps:
             - uses: actions/checkout@v5
             - name: Install `apt` packages
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,9 @@ jobs:
                 python-version: [
                     '3.10',
                     '3.11',
+                    '3.12',
+                    '3.13',
+                    '3.14',
                     ]
         steps:
             - uses: actions/checkout@v5
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ on:
 jobs:
     build:
         name: Build
-        runs-on: ubuntu-22.04
+        runs-on: ubuntu-24.04
         strategy:
             matrix:
                 python-version: [
@@ -68,7 +68,7 @@ jobs:
                 ./run_tests.py --outofsource full
     docs:
         name: Build documentation
-        runs-on: ubuntu-22.04
+        runs-on: ubuntu-24.04
         if: github.ref == 'refs/heads/main'
         strategy:
             matrix:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,9 +21,9 @@ jobs:
                     '3.11',
                     ]
         steps:
-            - uses: actions/checkout@v2
+            - uses: actions/checkout@v5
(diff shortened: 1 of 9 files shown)


johnyf:
The tests now pass. Previous test failures were due to a dependency intermittently not installing.

.github/workflows/main.yml
```diff
@@ -13,17 +13,20 @@ on:
 jobs:
     build:
         name: Build
-        runs-on: ubuntu-22.04
+        runs-on: ubuntu-24.04
```
slivingston, line 16: `ubuntu-22.04` is used in the `docs` job (line 71). We may want to use `ubuntu-24.04` there for consistency with this `build` job. However, it is acceptable to keep `ubuntu-22.04` for the `docs` job because Ubuntu 22.04 will be [officially supported for 18 months more](https://ubuntu.com/about/release-cycle), and I did not see a clear indication that `ubuntu-22.04` will be deprecated soon in the [runner-images repository](https://github.com/actions/runner-images).

If there is no argument against it, I recommend that we use `ubuntu-24.04` in the `docs` job, too.

contrib/xml/xmlio.py
```diff
@@ -47,7 +45,7 @@
 import xml.etree.ElementTree as ET
 
 
-_Data: _ty.TypeAlias = (
+type _Data = (
```
slivingston: [The `type` statement is new in Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html), so this change causes a syntax error in previous versions of Python. Because this is under contrib/, it is acceptable that the minimum version of Python to use xmlio.py is different than that of TuLiP (which is 3.10), but I wanted to comment on it to confirm that you are aware.

slivingston (approved):
Good work! I added 2 non-blocking comments.

.github/workflows/main.yml
johnyf: I agree, I added a commit that uses `ubuntu-24.04` for the `docs` job.

contrib/xml/xmlio.py
johnyf: To avoid the error in Python 3.10 and 3.11, I added a commit that removes this `type` statement. I was considering the possibility that code under `contrib/` could require a newer Python version to run, but thinking again about it, using the `type` statement is not necessary in this case.

johnyf:
Thanks for reviewing and approving the changes. I added fixup commits that:

- to build the documentation:
  - use Ubuntu 24.04
  - use Python 3.14 (was 3.10)
  - update the `checkout` and `setup-python` GitHub Actions

  I tested these changes on branch `ci_test_docs`.

- remove the `type` statement in `xmlio.py`
- fix issues found by type checking in `xmlio.py`

slivingston (approved):

PR #269, opened 2026-03-05
REL: update versions of dependencies

johnyf:
- In `setuptools == 82.0.0`, `pkg_resources` has been removed, thus requiring `omega >= 0.4.0`, which requires Python >= 3.11. `tulip` is updated to require Python >= 3.11 and `omega >= 0.4.0`, in order to avoid `pip` attempting to install `omega == 0.3.1` in combination with `setuptools == 82.0.0`, which are incompatible. Python 3.10 is omitted from CI environments.

- How `dd` is built in CI is updated, because `dd >= 0.6.0` requires Python >= 3.11, so it will be installed in all supported Python environments, which enables using environment variables in combination with `pip install` for `dd`. Previously, for Python 3.10 `dd == 0.5.7` was used, thus `python setup.py install` was still used.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,6 @@ jobs:
         strategy:
             matrix:
                 python-version: [
-                    '3.10',
                     '3.11',
                     '3.12',
                     '3.13',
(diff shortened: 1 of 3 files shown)


slivingston (approved):
I agree with increasing the minimum required Python version to 3.11. It still has 7 months remaining of [source-only security fixes](https://peps.python.org/pep-0619/#source-only-security-fix-releases), but it is [past the support window recommended by SPEC](https://scientific-python.org/specs/spec-0000/).

johnyf:
Thanks for reviewing and merging the changes.

